PR #82, opened 2015-07-03
Selectes radio button on first visit

azevedo-252:
The radio buttons don't select any option when we first arrive at it via the keyboard.
Now they do

### DIFF
--- a/source/javascripts/_input.es6
+++ b/source/javascripts/_input.es6
@@ -27,5 +27,17 @@ $(function() {
     });
   }
 
+  $.fn.radioify = function() {
+    let $elements = $(this);
+
+    $elements.on('focusin', function(e) {
+      let $radioButton = $(e.target);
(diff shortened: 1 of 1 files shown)


source/javascripts/_input.es6
```diff
@@ -27,5 +27,17 @@ $(function() {
     });
   }
 
+  $.fn.radioify = function() {
+    let $elements = $(this);
+
+    $elements.on('focusin', function(e) {
+      let $radioButton = $(e.target);
+      if (!$radioButton.filter(':checked').val()) {
```
zamith: What about `$radioButton.find(':checked').length == 0` or something like that? Not sure what reads better, or is more idiomatic with jQuery.

azevedo-252: Not sure either. I kind of went with the best answer I found.

zamith:
:+1:

PR #83, opened 2015-07-03
Fixes hero width

naps62:
On browsers with fixed scrollbars (which take up space), the hero images were adding a small horizontal scroll on smaller resolutions (<1100px)

PS: RatioEnforcer conflicts with u-onlySmall and u-exceptSmall, since both set the `display` property, so I had to separate them


azevedo-252:
:+1:

PR #85, opened 2015-07-03
Fixes company values on IE Mobile

naps62:
This is probably a fault in the svg's. IE seems to scale them
differently than other browsers, so when the SVG itself does not fully
specify how it should scale, problems happen (this is just my theory)

In this case, on IE mobile, the values panel was growing to about 400% width. And only one of the svgs was actually visble within it (even the text was being hidden somehow)

Moving everything to objects seems to fix it. And I believe it also makes more
sense from a performance perspective


azevedo-252:
:+1:

PR #87, opened 2015-07-06
Some a11y improvements

zamith:
Changes h1 to h2 inside a footer, because it is not yet supported by all
SRs.

### DIFF
--- a/source/community.slim
+++ b/source/community.slim
@@ -4,7 +4,7 @@ page: community
 ---
 
 .Panel.Panel--alternative
-  = partial 'nav'
+  .u-navPlaceholder
   = partial 'community/hero'
 = partial 'community/events'
 = partial 'community/open_source'
--- a/source/partials/_nav.slim
+++ b/source/partials/_nav.slim
@@ -5,8 +5,8 @@
   = nav_overlay_item('Apprenticeship', '/apprenticeship')
   = nav_overlay_item('Blog', '/blog')
 
-nav#main-nav.Nav.Nav--topPositioned class="#{current_page.data.nav_class}" aria-label=('Main navigation')
-  = link_to '/', class: 'Nav-logo'
+header#main-nav.Nav.Nav--topPositioned class="#{current_page.data.nav_class}"
+  = link_to '/', "aria-label" => "Homepage", class: 'Nav-logo'
     .NavLogo class=nav_logo_class
 
   .Nav-menu
@@ -15,10 +15,10 @@ nav#main-nav.Nav.Nav--topPositioned class="#{current_page.data.nav_class}" aria-
       span.Burger-line
       span.Burger-line
 
-  ul.Nav-items
-    = nav_item('Work', '/work')
-    = nav_item('Company', '/company')
-    = nav_item('Community', '/community')
-    = nav_item('Apprenticeship', '/apprenticeship')
-    = nav_item('Blog', '/blog')
-.u-navPlaceholder
+  nav aria-label='Main navigation'
+    ul.Nav-items
+      = nav_item('Work', '/work')
+      = nav_item('Company', '/company')
+      = nav_item('Community', '/community')
+      = nav_item('Apprenticeship', '/apprenticeship')
+      = nav_item('Blog', '/blog')
--- a/source/company.slim
+++ b/source/company.slim
@@ -4,7 +4,7 @@ page: company
 ---
 
 .Panel
-  = partial 'nav'
+  .u-navPlaceholder
   = partial 'company/hero'
 = partial 'company/history'
 = partial 'company/clients'
--- a/source/hire-us.slim
+++ b/source/hire-us.slim
@@ -3,7 +3,7 @@ title: Subvisual - Hire Us
 ---
 
 .Panel
-  = partial 'nav'
+  .u-navPlaceholder
   = partial 'hire-us/hero'
 = partial 'hire-us/form'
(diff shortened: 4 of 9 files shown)


source/community.slim
```diff
@@ -3,11 +3,13 @@ title: Subvisual - Community
...
 
-.Panel.Panel--alternative
-  = partial 'nav'
-  = partial 'community/hero'
-= partial 'community/events'
-= partial 'community/open_source'
-= partial 'community/we_are_present'
-= partial 'subvisual_separator'
-= partial 'footer'
+= partial 'nav'
```
azevedo-252: If it's outside of the main, what about moving this to the layout?

source/partials/_nav.slim
```diff
@@ -5,20 +5,21 @@
   = nav_overlay_item('Apprenticeship', '/apprenticeship')
   = nav_overlay_item('Blog', '/blog')
 
-nav#main-nav.Nav.Nav--topPositioned class="#{current_page.data.nav_class}" aria-label=('Main navigation')
-  = link_to '/', class: 'Nav-logo'
-    .NavLogo class=nav_logo_class
+header
+  #main-nav.Nav.Nav--topPositioned class="#{current_page.data.nav_class}"
```
azevedo-252: Why not attaching these classes and id to the <header> ?

source/partials/_nav.slim
```diff
@@ -5,20 +5,21 @@
...
+    = link_to '/', "aria-label" => "Homepage", class: 'Nav-logo'
+      .NavLogo class=nav_logo_class
 
-  .Nav-menu
-    .Burger class=nav_burger_class
-      span.Burger-line
-      span.Burger-line
-      span.Burger-line
+    .Nav-menu
+      .Burger class=nav_burger_class
```
azevedo-252: Maybe this could be a `<button>`?

source/partials/_nav.slim
```diff
@@ -5,20 +5,21 @@
...
+        span.Burger-line
 
-  ul.Nav-items
-    = nav_item('Work', '/work')
-    = nav_item('Company', '/company')
-    = nav_item('Community', '/community')
-    = nav_item('Apprenticeship', '/apprenticeship')
-    = nav_item('Blog', '/blog')
-.u-navPlaceholder
+    nav aria-label='Main navigation' role='navigation'
```
azevedo-252: I'm not a fan of having this element without a `Nav-*` class.
Couldn't this have the `.Nav-items` class? One thing I've been wondering and researched one time, is that a nav doesn't require its items or anything inside to be a `ul`. Is that *list* really necessary?

source/community.slim
zamith: Makes sense to me. :+1:

source/partials/_nav.slim
zamith: I just followed what is being done in the `footer`, thought it was some SUIT thing. :p

source/partials/_nav.slim
zamith: If you want a list of links, I'd say it is. What would be the alternative? To have a bunch of divs?

The thing is that the `Nav` css component regards the style of the whole thing, which in hindsight could have been called `Header`, or `TopBar`. The `nav` element has a very specific semantic purpose. I think its an unfortunate coincidence that they have the same name, but it should in no way affect how we construct the markup (apart from maybe renaming the class). Does this make sense to you?

source/partials/_nav.slim
zamith: This entire menu need a a11y revamp, so what about leaving this out of this PR?

source/partials/_nav.slim
azevedo-252: It's just that you'll need a 3 level deep structure:
`nav > ul > li`.
Many times had I the need to do some CSS extra work just to make the whole thing work due to the extra level.
Not sure the `list` brings anything semantically speaking since the nav is already self explanatory. That's my opinion, not a `de facto` answer.

source/partials/_nav.slim
azevedo-252: So what will you do?

source/partials/_nav.slim
zamith: It seems that having an `ul` inside of the `nav` is definitely the way to go. 

> Note: In cases where the content of a nav element represents a list of items, use list markup to aid understanding and navigation.

http://www.w3.org/TR/html5/sections.html#the-nav-element

source/partials/_nav.slim
zamith: I can change this, np.

source/partials/_nav.slim
azevedo-252: Hmm ok then

azevedo-252: But I still feel the same regarding the *unclassed* nav

zamith:
@azevedo-252 bump

source/partials/_nav.slim
zamith: Not sure what you want me to do here. I don't share your concerns regarding having an element that is purely semantic, without a class, because I think those are two completely different scopes.

Moving the `Nav-items` up would mean having to change the CSS.

azevedo-252: Mostly because I don't like letting things happen by chance.
But ok, we can go with this.

azevedo-252:
:+1:

PR #88, opened 2015-07-07
Styleguide working again

zamith:


azevedo-252:
The styleguide needs some curating now that everything is more stable.
:+1:

PR #90, opened 2015-07-07
Responsive images

azevedo-252:
Responsive images


naps62:
unless there's a reason for those images to have 2880px width, I'd resize them to 1920px

Other than that, :+1:

PR #91, opened 2015-07-07
Fix CS's case study

azevedo-252:
https://trello.com/c/PiAlcAKI/95-cs-case-study-review


source/stylesheets/case-studies/creators-school/_stripe.scss
```diff
@@ -29,6 +29,8 @@ $cs-Stripe-triangleHeight-mobile: 90px;
   &:after {
     content: '';
 
+    z-index: 500;
```
naps62: shouldn't this be adding to one of the existing layer variables instead of a hardcoded value?

azevedo-252: Sure

naps62:
:+1:

PR #92, opened 2015-07-07
Community hero animation

joaojusto:


source/community.slim
```diff
@@ -3,6 +3,10 @@ title: Subvisual - Community
 page: community
 ---
 
+- content_for :javascripts
+  = javascript_include_tag 'snap'
+  = javascript_include_tag 'community/all'
```
naps62: snap should go away.
and the rest should go to the main manifest

source/javascripts/community/all.js
```diff
@@ -0,0 +1,3 @@
+//= require gsap
```
naps62: do you need the full gsap library?

source/javascripts/community/hero_animation.js
```diff
@@ -0,0 +1,169 @@
...
+  function populateBuildings(building, index) {
+    buildings.push($(building.node));
+    populateWindows(building, index);
+  }
+
+  function populateWindows(building, index) {
+    var buildingWindow = building.select(buildingsNames[index].windows);
+    buildingWindow = buildingWindow.selectAll('g');
+    buildingWindow.items.reverse();
+    buildingWindows.push(buildingWindow);
```
naps62: `buildingWindow` is probably not the correct name, since this is a group

PR #93, opened 2015-07-07
Fixed button alignments

azevedo-252:
Created an hacky component, that only makes sense if combined and used in a certain way.
Basically it needs to have the columns' height and then the bottom element is absolutely positioned.
It might become less obvious how we should use margins when playing along with this component, since it positions the element outside the container

### DIFF
--- a/source/partials/community/_events.slim
+++ b/source/partials/community/_events.slim
@@ -56,8 +56,8 @@
 .Panel.Panel--alternative
   .HorizontalGrid.HorizontalGrid--limited
     #bragajs.HorizontalGrid-cell
-      .GridCell
-        .GridCell-content
+      .GridCell.u-flexGrow.u-defaultMargin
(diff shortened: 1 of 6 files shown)


source/partials/community/_events.slim
```diff
@@ -56,8 +56,8 @@
 .Panel.Panel--alternative
   .HorizontalGrid.HorizontalGrid--limited
     #bragajs.HorizontalGrid-cell
-      .GridCell
-        .GridCell-content
+      .GridCell.u-flexGrow.u-defaultMargin
+        .GridCell-content.StickToTheBottom.u-defaultThenLargeMargin
```
naps62: not to be too picky, but isn't the `The` part of the name unnecessary? `StickToBottom` would work just the same

azevedo-252: I'll change it

naps62:
:+1:

PR #95, opened 2015-07-08
Corrects the nav

zamith:
In order for the navbar in mobile to work correctly with the new
structure, we need a class on the actual `nav` element, so it can by
hidden in those resolutions.


azevedo-252:
Does this fix what currently happens with the nav button in mobile resolution?

azevedo-252:
:+1:

PR #102, opened 2015-08-07
Fix margins on prdocut development, consulting and training

joaojusto:
old:

screen shot 2015-08-07 at 17 20 52

new:

screen shot 2015-08-07 at 17 21 01


naps62:
:+1:

source/partials/consulting/_hero.slim
```diff
@@ -1,5 +1,8 @@
+.u-navPlaceholder
```
azevedo-252: This is not supposed to be used here I believe. It's used with the nav partial only

PR #103, opened 2015-08-19
Fixes styleguide

zamith:
Why:

* The styleguide was not showing the styles

* Stop requiring the font-awesome-sass, which we have stopped using a
while ago
* Ignoring some of the SASS special sintax in animations, which the
linter seems to dislike


azevedo-252:
:+1:

PR #104, opened 2015-08-24
Better ssl security

naps62:
. Inspired by https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html

Current results are now:
screenshot from 2015-08-24 15-53-17


zamith:
:+1:

PR #106, opened 2015-10-06
Fix HTML and SVG's

azevedo-252:
* Removed duplicate headers on almost every page
* Removed duplicate ids
* Cleaned useless id's
* Fixed invalid HTML


naps62:
:+1:

PR #111, opened 2015-10-30
Fix Creators School plug for Easy Money case study.

joaojusto:
Why:

* The Creators School plug for easy.money case study was not displaying the background correctly on Firefox.

* Removing the extra `Panel` component from the root of the template.
* Adding the `Panel--easymoneyGradient` modifier to the existing `Panel`.


zamith:
I think this is fine. /cc @azevedo-252

azevedo-252:
:+1:

PR #114, opened 2015-11-05
Two less HTTP requests

naps62:
Why:

* Bugherd was still being loaded, even though we no longer use it
* Segment was being loaded on a different request, instead of being
concatenated to the all.js manifest

* Removing Bugherd
* Moving segment to an erb template (to maintain the `build?` condition), and
loading it via the js manifest


zamith:
Does moving segment into the manifest improve the overall load page time?

naps62:
I'm not sure, since it's only loaded in production (and measuring locally would be pointless), but I have strong reasons to believe it will

zamith:
Well, we'll have to see then. :ship:

PR #117, opened 2015-12-10
Add Intercom

pfac:
Why:

* 3rd party integrations are to be loaded asynchronously, in such a way
that the impact on the page's rendering is minimized.

* Adding the supplied scripts to `vendor`
* Adding a new manifest file specifically for third party integrations
* Adding the new manifest file to the layout so it gets appended to all
pages


naps62:
you'll need to do this on https://github.com/subvisual/blog.subvisual.co and https://github.com/subvisual/apprenticeship.subvisual.co as well

pfac:
On it

source/layouts/layout.slim
```diff
@@ -57,3 +57,8 @@ html.Theme itemscope='' itemtype='http://schema.org/WebPage' lang=data.site.lang
       try{Typekit.load();}catch(e){}
     = javascript_include_tag 'all'
     = yield_content :javascripts
+    javascript:
```
zamith: What about doing an `unless development?` here?

naps62: actually `if production?` would be better. we don't want this in test/staging either

naps62:
couldn't this be inside the manifest with the rest of the JS code? I moved some stuff there recently precisely to avoid the extra HTTP request that this will cost us

gabrielpoca:
@naps62 in this case maybe we can benefit from the fact that this file can stay cached forever instead of merging with the existing JS.

zamith:
@gabrielpoca What do you mean by forever? And are you talking about browser cache?

gabrielpoca:
I mean that the file will never change, when our js will, the browser can cache it longer, and maybe we can even use service works or appcache to keep it forever.

zamith:
My browser cache last about 5 days, let's say that the average user has a cache of 15 days, in practice that means that he'll probably never have files from our site cached, and if he does, it's as likely that he has `application.js` as it is for `intercom.js`.

Appcache is a possibility, but I don't think this amount of code justifies that. If we can get everything in one http call, that should be fast enough. We can test it though, with and without app cache.

@naps62 One thing we have never used is an reverse proxy such as varnish, would that improve the load times substancially?

zamith:
@gabrielpoca Actually after some research I found that appcache has been deprecated in favor of service workers, which are not supported at all in IE or Safari.

Also, if I'm not mistaken you need even more JS just to make that work, so not really a good option I would say, but not very familiar with it I have to say.

gabrielpoca:
@zamith it was just an idea, I said service works or appcache, or maybe a combination of both. I'm eager to see service workers in action, but I guess it's not the best place.

pfac:
I would go with having it on the manifest, as the scripts are not updated with a regularity that justifies it ([4 commits in 6 months](https://github.com/subvisual/subvisual.co/commits/master/source/javascripts).

Here's my reasons for having done it like so:

1. The fact that Intercom provided this code as two separate `<script>` tags leads me into thinking they have dealt with some sync issue in the past and want to prevent some race condition;
2. Intercom is an addon to the site, so it makes more sense to load it after all scripts required by these pages have been loaded, which means this would have to be done after `yield_content :javascripts`
3. I wasn't able to have Intercom run successfully in development (also didn't dig into the rabbit hole tbh), which took the confidence to try and challenge the way they prepared this code.

Now, if you guys these don't have a meaningful impact, then I'd rather have it in the manifest file and merge the Intercom configuration in the same `<script>` tag as Typekit.

source/layouts/layout.slim
pfac: I'm perfectly ok with it.

naps62:
I was just following the same rationale that I applied here: https://github.com/subvisual/subvisual.co/pull/114

We might be fine either way, but since we already moved other external services to the manifest, I'd say it's better to keep things consistent.

Assuming the file can stay cached forever is a mistake, not only because of what zamith said, but because of what's stated in [this post](http://www.nateberkopec.com/2015/11/05/page-weight-doesnt-matter.html), section "HTTP Caching"

naps62:
@zamith as for varnish (or a similar tool), it's still something I have on my backlog titled "DevOps things to try out when I have the time"

naps62:
@pfac Also, having two separate script tags is probably irrelevant. They probably set it up like that so that you can use a CDN for the main script

gabrielpoca:
I was thinking about this and I believe we should load intercom asynchronously. It will not change the page weight. It will make an extra request, but in the other hand the JS that is crucial to the site will be available faster.

zamith:
@gabrielpoca I wouldn't say it is crucial. I'd rather people can see the page faster and intercom appear later, than having them wait for it. It is important for us to have intercom there, but it is not crucial that it is there at the time of rendering.

gabrielpoca:
@zamith apart from that, I didn't understand if you agree with my idea or not.

zamith:
@gabrielpoca What I'm saying is that I don't really care when it is loaded, it can be 1s or 2s after the page being rendered, as long as it has no impact on the rendering time.

gabrielpoca:
@zamith but it does affect. Because our JS is necessary for things like the Nav to work.

zamith:
@gabrielpoca Ok, then why not have a `defer` for intercom?

gabrielpoca:
@zamith I don't understand. I suggested loading is asynchronously, aren't you talking about the same thing?

zamith:
@gabrielpoca As far as I understand it, `async` will stop render to execute the script whenever it is ready, and `defer` only executes it after render, which is more in sync with what I was saying. Is this not true?

gabrielpoca:
I guess, I don't know, I didn't suggest `async` or `defer`, I simply suggested asynchronously. If that information is correct, then `defer` makes more sense.

zamith:
Oh, I thought I had read `async`, my bad. Yes, I think `defer` is the way to go.

naps62:
Again, quoting from [this same post](http://www.nateberkopec.com/2015/11/05/page-weight-doesnt-matter.html):

> It's far better to use async and defer attributes on your script tags

to clarify, he states that we should use both. And he talks about this in precisely the context of loading 3rd party integrations

gabrielpoca:
Then I guess we all agree on what to do :tada:

zamith:
I'm good with that, but I'd like to understand why you would use both. I don't think he explains it in that post. :/

zamith:
I think this is it:

> The defer attribute may be specified even if the async attribute is specified, to cause legacy Web browsers that only support defer (and not async) to fall back to the defer behaviour instead of the blocking behaviour that is the default.

found [here](http://www.w3.org/html/wg/drafts/html/master/single-page.html#attr-script-async).

pfac:
\o/ nice! Will change Intercom to the vendor dir and load it with both `defer` and `async`.

pfac:
Changed

source/layouts/layout.slim
```diff
@@ -57,3 +57,9 @@ html.Theme itemscope='' itemtype='http://schema.org/WebPage' lang=data.site.lang
       try{Typekit.load();}catch(e){}
     = javascript_include_tag 'all'
     = yield_content :javascripts
+    - unless development?
```
naps62: did you test this? I don't think `development?` is really a thing. @zamith wrote that as pseudocode I think

We've been using `if build?` so far
example: https://github.com/subvisual/subvisual.co/blob/b4d53f1221e367c9489299ed35490333234efe12/source/javascripts/_segment.js.erb

zamith: It is an actual thing:
https://github.com/middleman/middleman/blob/v3.0.6/middleman-core/lib/middleman-core/application.rb#L200-206

On Fri, Dec 11, 2015 at 4:57 PM Miguel Palhas <notifications@github.com> wrote:
In source/layouts/layout.slim :

> @@ -57,3 +57,9 @@ html.Theme itemscope='' itemtype='http://schema.org/WebPage' lang=data.site.lang
>        try{Typekit.load();}catch(e){}
>      = javascript_include_tag 'all'
>      = yield_content :javascripts
> +    - unless development?


did you test this? I don't think development? is really a thing. @zamith wrote that as pseudocode I think

We've been using if build? so far
example: https://github.com/subvisual/ subvisual.co/blob/ b4d53f1221e367c9489299ed354903
33234efe12/source/javascripts/ _segment.js.erb

pfac: @naps62 `bin/server`:
screen shot 2015-12-11 at 18 25 13

`middleman server -p 3000 -e production`:
screen shot 2015-12-11 at 18 28 48

naps62: didn't know that

:+1: :shipit:

pfac:
:tada:

PR #119, opened 2016-01-20
Hide hellobar on small devices

gabrielpoca:
Why:

The pop-up is really annoying on mobile devices. This hides it.


joaojusto:
:+1:

PR #122, opened 2016-03-11
Refactor Nav

azevedo-252:
Depends on: https://github.com/subvisual/blue/pull/98

Why:

* The apprenticeship link should be replaced by a button linking to the
/hire-us page

screen shot 2016-03-11 at 12 16 30 pm

### DIFF
--- a/source/partials/_nav.slim
+++ b/source/partials/_nav.slim
@@ -2,8 +2,8 @@
   = nav_overlay_item('Work', '/work')
   = nav_overlay_item('Company', '/company')
   = nav_overlay_item('Community', '/community')
-  = nav_overlay_item('Apprenticeship', '/apprenticeship')
   = nav_overlay_item('Blog', '/blog')
+  = nav_overlay_item('Work with us', '/hire-us', '', { link_class: 'ButtonSecondary' })
 
 header#main-nav.Nav.Nav--topPositioned class="#{current_page.data.nav_class}"
   = link_to '/', "aria-label" => "Homepage", class: 'Nav-logo'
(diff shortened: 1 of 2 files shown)


source/partials/_nav.slim
```diff
@@ -2,8 +2,9 @@
   = nav_overlay_item('Work', '/work')
   = nav_overlay_item('Company', '/company')
   = nav_overlay_item('Community', '/community')
-  = nav_overlay_item('Apprenticeship', '/apprenticeship')
   = nav_overlay_item('Blog', '/blog')
+  li.NavOverlay-item
```
zamith: Why can't this use the `nav_overlay_item` helper?

source/partials/_nav.slim
```diff
@@ -20,5 +21,6 @@ header#main-nav.Nav.Nav--topPositioned class="#{current_page.data.nav_class}"
       = nav_item('Work', '/work')
       = nav_item('Company', '/company')
       = nav_item('Community', '/community')
-      = nav_item('Apprenticeship', '/apprenticeship')
       = nav_item('Blog', '/blog')
+      li.Nav-itemNoUnderline
```
zamith: Could this be an option to `nav_item`?

source/partials/_nav.slim
azevedo-252: I've tried it. Too complex to acgieve such a simple and yet different output

zamith: Ok.

source/partials/_nav.slim
azevedo-252: I guess it can

source/partials/_nav.slim
azevedo-252: Actually I can't say why I thought It was difficult. Fixed using the helper

zamith:
👍

PR #124, opened 2016-03-11
Replace homepage hero

azevedo-252:
Note: there's a breakpoint range that's not yet taken care of, because it's waiting for a new image from the @subvisual/designers 

Replace homepage hero

Why:

* The homepage's hero is being refactored. Main changes:
  * New copy
  * New hero shot
* The SVG animations that are being loaded via `<object>` are not
triggering sometimes because of a caveat of the `load` event from JQuery
when used with `images`/`objects`:
  * trying to execute a function when an image (or collection of images) has
  completely loaded can cease to fire for images that already live in the
  browser's cache
* There's no perfect solution for this, so we came around with a
solution which uses `setInterval` and searches (repeatedly) the *image*
in the DOM and then starts the animation when it finds it, instead of
using `$.load`

* Refactoring the planets image (SVG) to a simpler version without
lines
* Replacing every `$.load` that were handling SVG's loading for our
custom event handler
* Refactoring the way we load the SVG and trigger its animation by
dynamically setting the SVG's data, and only after the load event
triggers we start the animation
* Refactoring the hero parallax' script:
  * uses more ES6 syntax
  * remove *lines* as a target of the animation
  * expose a single function to start the animation instead of auto
  starting when the script loads
* Refactoring the lines animation script:
  * uses more ES6 syntax
* Refactor the hero's markup to use the new copy and style (mainly
layout)
* Adding a custom CSS component for dealing with the homepage's hero
copy style
* Adding a new modifier to `.Panel` to manage the hero's layout
responsively
* Adding a new `.RatioEnforcer` modifier to enforce a max-width to the
background
* Adding a new group of utilities specific to the website's ad-hoc
breakpoints ;)

screen shot 2016-03-11 at 12 16 30 pm

screen shot 2016-03-11 at 12 53 00 pm


source/index.slim
```diff
@@ -20,7 +20,13 @@ nav_logo_class: 'is-animated'
...
   javascript:
-    $('#HomepageHero').load(function() {
-      var svg = $('#HomepageHero').contents().find('svg')[0];
-      window.homepage.animateLines(svg);
-    })
+    var heroWrapper = $('#HomepageHero');
+    var objectElement = $('#HomepageHeroSVG');
+
+    objectElement
+      .on('load', function() {
```
zamith: Is it possible to keep the `.` in the previous line? We do that in ruby because we feel it's easier to understand that it's part of the previous line that way.

source/index.slim
```diff
@@ -20,7 +20,13 @@ nav_logo_class: 'is-animated'
...
+        var svg = objectElement.contents().find('svg')[0];
+
+        window.homepage.parallax(svg, heroWrapper[0]);
+      })
+      .attr('data', objectElement.attr('data'));
```
zamith: Does this run before `load` is triggered?

source/partials/homepage/_hero.slim
```diff
@@ -1,18 +1,26 @@
...
+        .HomepageHero-copy
+          h1.HomepageHero-heading.u-xSmallThenSmallMargin
+            ' We
+            span.HomepageHero-headingHighlight> design
+            ' and
+            span.HomepageHero-headingHighlight> develop
+            ' thoughtful digital
+            span.HomepageHero-headingHighlight products
+          p.HomepageHero-text.u-defaultMargin
+            | We craft products with a focus on great User Experience and high-quality code
```
zamith: About this copy, what about going more in the direction of what the client can get from us and not what we can give to the client, if that makes sense. /cc @johnny1337

naps62:
@azevedo-252 can you deploy this to the test server, so we can preview it?

`DEPLOY_BRANCH=new-hero bundle exec cap staging deploy`

source/index.slim
azevedo-252: There's no guarantee of that.
This exists as an hack that should have been explained in the commit message.
I'll go with a different approach and update the commit message properly

azevedo-252:
@naps62 @zamith bump
I've changed a few things, so it's better for you to review everything again

source/index.slim
```diff
@@ -20,7 +20,15 @@ nav_logo_class: 'is-animated'
...
 - content_for :javascripts
   javascript:
-    $('#HomepageHero').load(function() {
-      var svg = $('#HomepageHero').contents().find('svg')[0];
-      window.homepage.animateLines(svg);
-    })
+    var heroWrapper = $('#HomepageHero')[0];
+    var objectElement = $('#HomepageHeroSVG');
+
+    function svgFinder() {
```
naps62: is it just me, or is this method's name confusing?

without looking at it's content, I'm unable to tell whether `svgFinder()` would return the found svg, or a callback that would then find the svg

Not sure if I explained myself properly...

source/javascripts/_custom_events.js
```diff
@@ -0,0 +1,13 @@
+(function(window) {
+  function onLoad(svgFinder, callback) {
```
naps62: svgFinder looks extremely arbitrary as the argument name here. Isn't this suposed to be a more generic thing?

source/javascripts/homepage/_logo_tools.es6
```diff
@@ -1,61 +1,69 @@
...
+      window.animationHelpers.onScrollToElement('#LogoTools', 1, animateAllTools);
+    }, 500);
+  }
 
-  setTimeout(function() {
-    window.animationHelpers.onScrollToElement('#LogoTools', 1, animateAllTools);
-  }, 500);
-});
+  window.customEvents.onLoad(svgFinder, startAnimation);
+})(window);
```
naps62: since you touched this file, why not `s/var/let/`?

source/stylesheets/utilities/_viewport.scss
```diff
@@ -0,0 +1,68 @@
...
+// markup:
+// <div class="u-onlyAboveResponsibleResponsiveDesignPicture">I only show above a certain resolution</div>
+//
+// Styleguide 5.18.4
+
+.u-onlyAboveResponsibleResponsiveDesignPicture {
+  @include media('<=1035px') {
+    display: none;
+  }
+}
```
naps62: I can't find a reference to any of these classes in the PR. Where they extracted from blue?

source/partials/homepage/_hero.slim
```diff
@@ -1,18 +1,26 @@
...
-.u-smallMargin
-.GridCell.GridCell--alignCenter.u-defaultThenDoubleLargeMargin
-  .GridCell-content
-    .ContentFormatter.ContentFormatter--centerHorizontally
-      .u-smallThenDefaultMargin
-      p.Text.Text--dark
-        | We handcraft thoughtful digital experiences that empower individuals, institutions and communities
+#HomepageHero.Panel.Panel--homepage
+  .Panel-content
+    .HomepageHero
```
naps62: it looks confusing to me that `#HomepageHero` and `.HomepageHero` are actually two different elements

naps62:
@azevedo-252 I just deployed this to test.subvisual.co

Isn't the hero parallax effect reversed now? It feels weird to me like this

rmdmachado:
Bump!

azevedo-252:
@rmdmachado Whenever I have the opportunity

rmdmachado:
@azevedo-252 when do you think that will be? Can I help with anything? Should we plan on having it done at the beginning or end of a day this week? Thanks.

source/index.slim
azevedo-252: True. I'll change to `findSVG`

source/javascripts/_custom_events.js
azevedo-252: True. I'll do this
```
function onLoad(callback, predicate = true) {
  const timerId = setInterval(function() {
    if (predicate) {
      callback();
      clearInterval(timerId);
    }
  }, 100);
}
```

source/javascripts/homepage/_logo_tools.es6
azevedo-252: Because a *var* means it hasn't been "polished". While if I see a `let` I will think it's a mutable, which it may not be if I apply that Find&Replace

source/stylesheets/utilities/_viewport.scss
azevedo-252: Yes. Fixed

source/partials/homepage/_hero.slim
azevedo-252: True. I'll change the *id* to `HomepageHeroWrapper`

source/javascripts/homepage/_logo_tools.es6
naps62: I'm fine with that

source/javascripts/_custom_events.js
naps62: :+1:

naps62:
:+1: on my part. no sure if @zamith has anything left to review

PR #127, opened 2016-06-21
Fix iOS issues with the hire us form

zamith:
Remove extra border in radio buttons on iOS

    Why:

    * We are styling radio buttons ourselves, on the `::before` pseudo element, so
      any extra style should be remove. Apparently there was an extra square border
      around our round radio button.

    * Forcing the border on the actual input to be invisible

Fix issue with hire us form overflowing

    Why:

    * In Safari on small dimensions (and by extension iOS), the hire us form was
      overflowing to the next panel. This was happening due to an issue with how
      Safari uses flex, when compared to webkit.

    * Only setting the `flex-grow` to 1 when the direction of flex is row, which is
      really the only place we need it and does not mess up with Safari

### DIFF
--- a/source/stylesheets/components/_hire_us_grid.scss
+++ b/source/stylesheets/components/_hire_us_grid.scss
@@ -23,10 +23,6 @@
   padding-right: $Theme-spacing-small;
 }
 
-.HireUsGrid-greetings {
-  flex: 1;
-}
-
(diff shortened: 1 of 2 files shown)


zamith:
@naps62 Can you check this?

naps62:
:+1:

PR #128, opened 2016-08-08
Update copyright year

pfac:
Why:

* It's already August, this should have been done already.

* Refactoring the pages footer to mention 2016 in the copyright text.

### DIFF
--- a/source/partials/_footer.slim
+++ b/source/partials/_footer.slim
@@ -50,7 +50,8 @@ footer
     .Footer-credits
       span.Footer-handcrafted Handcrafted by
       span.Footer-subvisual SUBVISUAL
-      span.Footer-copyright © 2016 Copyright by Subvisual
+      span.Footer-copyright
+        | © #{Date.today.year} Copyright by Subvisual
     = link_to '/hire-us', class: 'Footer-work'
       span.Footer-interested Interested in our services?
       .Footer-workText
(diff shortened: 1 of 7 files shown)


source/partials/_footer.slim
```diff
@@ -50,7 +50,7 @@ footer
     .Footer-credits
       span.Footer-handcrafted Handcrafted by
       span.Footer-subvisual SUBVISUAL
-      span.Footer-copyright © 2015 Copyright by Subvisual
+      span.Footer-copyright © 2016 Copyright by Subvisual
```
naps62: how about making this dynamic, so that we only need to trigger a deploy next year, instead of going over this once again?

pfac: Seems like a good idea to me. Refactored

naps62:
:+1:

zamith:
This will only update when we redeploy the site, right? Since this is generate on build time by middleman, or am I wrong?

naps62:
@zamith yes, that's exactly what I suggested in my comment earlier. It's sligthly better than having to manually change the year

zamith:
Sure. Just checking. Another possibility would be to have this as a default
and changing it with JS after the page is loaded. So it would always show
something that is probably right, even without JS.

But maybe it's too much work for such a small thing.

PR #129, opened 2016-08-09
Update PT office address

pfac:


source/partials/homepage/_contact.slim
```diff
@@ -23,4 +23,4 @@
         )
       .ContactsGrid-address
         p.Text.Text--light.Text--strong
-          ' Centro Empresarial de Braga
+          ' Pólo de Negócios de Braga
```
naps62: shouldn't we have the door number as well?

pfac: Since we didn't have it before, I don't see the reasoning to have the door number now. Also, this was never the complete office address, I guess that will be solved in the upcoming redesign

naps62:
:+1:

PR #130, opened 2016-08-11
Fix project setup script

pfac:
Why:

* It does not specify which executable should run the script;
* It is using NPM to install bower manually instead of using a package
  manifest file;
* It is installing Bower without the global option, but assuming that
  the Bower executable will be available globally;
* It is installing Node dependencies after Bower dependencies.

These changes address the issues by:

* Refactoring the setup script;
* Adding Bower as a dependency to the package manifest file.


naps62:
:+1:

PR #131, opened 2016-08-11
Replace links to HelloHired.com

pfac:
Why:

* HelloHired.com is no more.

* Refactoring the site configuration to replace links to hellohired.com
  with links to send us an e-mail.


naps62:
was this change approved by the board of directors?

rmdmachado:
Yeah! Unfortunately we need to change to a different service or implement a new approach, since right now the links are broken, it's just a quick temporary fix.

naps62:
:+1:

PR #132, opened 2016-08-16
Remove sections from community and work

gabrielpoca:
Why:

This is the first PR towards redesigning our website. In this change, I'm removing parts from the "Community" and "Work" pages.

* Removing and re-ordering sections.


naps62:
From what I could see :+1: 

But maybe it's better if someone else also reviews this

PR #134, opened 2016-08-23
Add skills section to homepage

gabrielpoca:
Why:

This change will add a section with skills, right after the hero.

* Adding icons for each skill.
* Adding markup.


gabrielpoca:
@naps62 updated

naps62:
:+1:

PR #135, opened 2016-08-24
Add section with blog posts

gabrielpoca:
Why:

In this change, I'm adding a section with the most recent blog posts to
the homepage. Because there is no back-end for this website, we decided to load the blog posts dynamically from an endpoint I recently created on the blog.

You'll notice a new component named `SectionHeading`. It uses custom font-sizes because this component is intended for aesthetics, not content. The font-size works the same way a `height` would on any other component. For that reason, I didn't make an effort to include those sizes in the system.

* Add a section for blog posts.
* Load blog posts from API and use a template to display on the screen.

### DIFF
--- a/source/javascripts/homepage/_blog_posts.es6
+++ b/source/javascripts/homepage/_blog_posts.es6
@@ -1,0 +1,36 @@
+(function(window, $) {
+
+  function loadBlogPosts() {
+    const xhttp = new XMLHttpRequest();
+    xhttp.onreadystatechange = function() {
+      if (xhttp.readyState == 4 && xhttp.status == 200)
+        updatePosts(JSON.parse(xhttp.responseText));
+    };
+    xhttp.open("GET", "https://subvisual.co/blog/posts.json", true);
+    xhttp.send();
+
+    function updatePosts(posts) {
+      posts.splice(0, 2).forEach(post => {
+        const author = post.author;
+        const template = $('#blog-post-template').clone().find('.HorizontalGrid-cell');
+        template.find('.blog-post-title').html(post.title);
+        template.find('.blog-post-content').html(post.processed_intro);
+        template.find('.blog-post-author-picture').attr('src', 'https://subvisual.co/' + author.photo_url);
+        template.find('.blog-post-author-name').attr('href', `https://twitter.com/${author.twitter_handle}`);
+        template.find('.blog-post-author-name').html(`${author.first_name} ${author.last_name}`);
+
+        const tags = template.find('.blog-post-tags');
+        post.tag_list.forEach((tag, index) => {
+          if (index !== 0) tags.append(', ');
+          const el = $(`<span class="Tag Tag--${tag}">${tag}<span>`);
+          tags.append(el);
+        });
+
+        $('#blog-posts').append(template);
+      });
+    }
+  }
+
+  window.homepage = window.homepage || {};
+  window.homepage.loadBlogPosts = loadBlogPosts;
+})(window, window.$);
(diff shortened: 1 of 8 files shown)


source/javascripts/homepage/_blog_posts.es6
```diff
@@ -0,0 +1,36 @@
+(function(window, $) {
+
+  function loadBlogPosts() {
+    const xhttp = new XMLHttpRequest();
+    xhttp.onreadystatechange = function() {
+      if (xhttp.readyState == 4 && xhttp.status == 200)
+        updatePosts(JSON.parse(xhttp.responseText));
+    };
+    xhttp.open("GET", "http://localhost:3000/blog/posts.json", true);
```
naps62: localhost :warning:

source/javascripts/homepage/_blog_posts.es6
```diff
@@ -0,0 +1,36 @@
...
+    xhttp.send();
+
+    function updatePosts(posts) {
+      posts.splice(0, 2).forEach(post => {
+        const author = post.author;
+        const template = $('#blog-post-template').clone().find('.HorizontalGrid-cell');
+        template.find('.blog-post-title').html(post.title);
+        template.find('.blog-post-content').html(post.body);
```
naps62: now that I see this, I realize you named the `processed_intro` attribute to `body` in the API.

I'm very much not a fan of that. there's already a `body` field in the post, which is not the one that's being sent, and can cause confusion

You mind if I change it back in the blog, which would require changing it here as well?

gabrielpoca: No, I agree with you. I'll open a PR.

source/javascripts/homepage/_blog_posts.es6
gabrielpoca: ups, updated

gabrielpoca:
@naps62 updated

source/javascripts/homepage/_blog_posts.es6
```diff
@@ -0,0 +1,36 @@
+(function(window, $) {
+
+  function loadBlogPosts() {
+    const xhttp = new XMLHttpRequest();
+    xhttp.onreadystatechange = function() {
+      if (xhttp.readyState == 4 && xhttp.status == 200)
+        updatePosts(JSON.parse(xhttp.responseText));
+    };
+    xhttp.open("GET", "http://subvisual.co/blog/posts.json", true);
```
naps62: you migth as well change this to `https` to avoid having a redirect all the time

source/partials/homepage/_blog_posts.slim
```diff
@@ -0,0 +1,19 @@
...
+    h2.Heading.Heading--blog.u-smallMargin.blog-post-title
+    p.blog-post-content
+    .u-xSmallMargin
+    .blog-post-tags
+    .u-smallMargin
+    .PostAuthor
+      img.PostAuthor-picture.blog-post-author-picture
+      .PostAuthor-name
+        span> by
+        a.Link.blog-post-author-name(href="#" target="_blank")
```
naps62: why is this a link?

gabrielpoca: It's replaced with the link for twitter.

source/javascripts/homepage/_blog_posts.es6
gabrielpoca: yes, makes sense.

naps62:
:+1:

PR #136, opened 2016-08-24
Remove unused section from homepage

gabrielpoca:
Why:

With the redesign lots of sections will be remove. This change removes
everything that's not necessary anymore on the homepage.

* Removing unused sections.


naps62:
:confetti_ball: :shipit:

PR #138, opened 2016-08-31
Add case studies to homepage

gabrielpoca:
Why:

The purpose of this change is to add a generic case study component element to the homepage. This component is responsive and can be used for different case studies.

* Creating a component.
* Adding images.
* Adding a section to homepage.


source/stylesheets/components/_case_study.scss
```diff
@@ -0,0 +1,125 @@
+// CaseStudy
+//
+// A compoent to showcase a project
+//
+// .CaseStudy--articles - Uses oncostats' background color
+// .CaseStudy--oncostats - Users articles' background color
```
zamith: I believe these are reversed.

source/stylesheets/components/_case_study.scss
```diff
@@ -0,0 +1,125 @@
...
+.CaseStudy {
+  position: relative;
+
+  display: flex;
+  flex-direction: column;
+
+  max-width: $Theme-grid-maxWidth;
+  margin: 0 auto;
+
+  @include media('>smallDesktop') {
```
zamith: With the increasing ubiquity of the web and the increasing number of different device sizes, shapes and types, would it make sense to stop using these aliases and use only the pixel value? The reasoning being that the style of `smallDesktop` might actually be show on a large tablet, or small tv, or something else, which is not inherently wrong, but kind of defeats the abstraction on which the alias was based in the first place.

I know this is very meta and I don't think this is a blocker to the merge of the PR, but it is something I have been wondering and felt this would be as good a place as any to start the conversation.

/cc @subvisual/css

gabrielpoca: I see what you mean; we could define a set of acceptable numbers for the media queries and use them instead of this alias. But I think that naming still helps to know what screen size we are targeting; maybe there is a naming system that isn't coupled to the screen type?

source/stylesheets/components/_case_study.scss
gabrielpoca: 👍 fixed

gabrielpoca:
@zamith bump

source/stylesheets/components/_case_study.scss
zamith: Yes, I think that would be better.

zamith:
👍

PR #160, opened 2017-01-26
Adds Full Story

frmendes:


source/layouts/layout.slim
```diff
@@ -15,6 +15,7 @@ html.Theme itemscope='' itemtype='http://schema.org/WebPage' lang=data.site.lang
 
     = yield_content :stylesheets
 
+    = partial "full_story"
```
naps62: does this really need to go here? (at the beginning of the document, and executed synchronously)

it goes pretty much against all the optimizations we've done recently

PR #161, opened 2017-03-03
Support more browsers

gabrielpoca:
Using BrowserStack we noticed that our landing page didn't look right in
many browsers. The main reason for that is we were only using
auto-prefixer for ie, while Firefox and Chrome still needs prefixing for
rules like `column-count`.

Another common issue is images taking more space than they are supposed
to. That was solved by limiting them to their parents with `width:
100%`. There was already a `max-width` in place, but it was not having
any effect.


naps62:
shipit

PR #165, opened 2017-06-14
Refactor nginx configs

naps62:
These will soon be removed after migrating to S3, but I need to refactor some things to make the transition easier

- Removes nginx config file from the source directory. They had no business being there in the first place
- Separates `groupbuddies.com` rules into it's own file.
- Updates SSL config on blog, to match the main site
- Simplifies some of the other domains, such as `test.subvisual.co` which doesn't need any fancy stuff such as www-redirecting, given that it is for internal use only

PS: These are already being used in production, and working properly (apparently)

### DIFF
--- a/nginx/subvisual.co.conf
+++ b/nginx/subvisual.co.conf
@@ -1,5 +1,10 @@
+#
+# main server
+#
 server {
-  listen 443; server_name subvisual.co;
+  listen 443 ssl;
+  server_name subvisual.co;
+
   root /apps/subvisual.co/current;
 
   error_page 404 = /404/;
@@ -12,9 +17,6 @@ server {
     auth_basic off;
   }
 
-  location ^~ /newsletter/nginx.conf {
-    return 404;
-  }
   location ^~ /nginx.production.conf {
     return 404;
   }
@@ -52,12 +54,6 @@ server {
     proxy_pass http://blog-subvisual-co;
   }
 
-  location ^~ /newsletter {
-    alias /apps/newsletter.subvisual.co/;
-    access_log /apps/newsletter.subvisual.co/log/nginx.out;
-    error_log  /apps/newsletter.subvisual.co/log/nginx.err;
-  }
-
   location / {
     index index.html;
   }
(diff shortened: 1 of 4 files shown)


nginx/subvisual.co.conf
```diff
@@ -1,5 +1,6 @@
 server {
-  listen 443; server_name subvisual.co;
+  listen 443;
+  server_name subvisual.co www.subvisual.co old.subvisual.co;
```
pfac: Shouldn't this also be recognizing the `.info` and `.pt` domain versions (including the `www` subdomain)?

nginx/subvisual.co.conf
```diff
@@ -78,43 +70,10 @@ server {
...
-  ssl_ciphers 'EECDH+AESGCM:EDH+AESGCM:ECDHE-RSA-AES128-GCM-SHA256:AES256+EECDH:DHE-RSA-AES128-GCM-SHA256:AES256+EDH:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA:ECDHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES128-SHA256:DHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES256-GCM-SHA384:AES128-GCM-SHA256:AES256-SHA256:AES128-SHA256:AES256-SHA:AES128-SHA:DES-CBC3-SHA:HIGH:!aNULL:!eNULL:!EXPORT:!DES:!MD5:!PSK:!RC4';
-  ssl_dhparam /etc/nginx/ssl/dhsecure.pem;
-  add_header Strict-Transport-Security "max-age=31536000; includeSubdomains;";
-}
-
 # http to https redirect
 server {
   listen 80;
-  server_name subvisual.co subvisual.pt subvisual.info;
+  server_name subvisual.co subvisual.pt subvisual.info www.subvisual.pt;
```
pfac: Same, here, shouldn't this one support all the `www` subdomains?

nginx/subvisual.co.conf
naps62: we don't have ssl certificates for those, so no.

naps62:
I'm cancelling this because apparently some redirects are not working. I might come up with  another version later

PR #169, opened 2017-06-19
Deploying to S3

naps62:
As of today, this is our new hosting setup for our website

So, in a sense, this PR is already tested


package.json
```diff
@@ -17,5 +17,8 @@
   "homepage": "https://github.com/groupbuddies/subvisual.co",
   "devDependencies": {
     "critical": "^0.8.1"
+  },
+  "scripts": {
+    "build": "gatsby build"
```
pfac: :warning:

PR #170, opened 2017-06-22
Add Norte 2020 logos

pfac:
Why:

* https://trello.com/c/q6zVNoL1

* Adding the Norte 2020 banner with the logos to the footer, between the
  links and the final section.


naps62:
@pfac the trello card mentions a hyperlink. where is that?

pfac:
@naps62 Forgot that, nice catch :+1:. Added

PR #178, opened 2018-02-09
Add Helena

pfac:
Why:

* https://trello.com/c/5HUhw7vp

* Adding Helena's photo;
* Adding Helena's info to the team file.

### DIFF
--- a/data/team.yml
+++ b/data/team.yml
@@ -35,6 +35,13 @@ members:
       twitter: "https://twitter.com/gabrielgpoca"
       github: "https://github.com/gabrielpoca"
 
+  - name: Helena Muniz
+    picture: helena-muniz
+    position: Head of People Operations
+    bio: >
+      Loves people, learning, and to travel. Thrives in helping the team shoot
+      for the moon. Believes everyone can achieve their goals and dreams. 
+
   - name: "João Ferreira"
     picture: "joao-ferreira"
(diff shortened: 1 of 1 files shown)


data/team.yml
```diff
@@ -35,6 +35,13 @@ members:
       twitter: "https://twitter.com/gabrielgpoca"
       github: "https://github.com/gabrielpoca"
 
+  - name: Helena Muniz
+    picture: helena-muniz
+    position: Head of People Operations
+    bio: >
+      Loves to travel, learning and people. Thrives in helping the team shoot
+      for the moon. Believes everyone can achieve their goals and dreams.
```
fribmendes: Shouldn't this be "anyone"?

pfac: Don't think so. The goal here is to imply that every person has the ability to achieve their goals. Anyone implies that some people do, who those people are is just unknown.

fribmendes: But that's the issue. When you say `Believes everyone can` you are referring to a set group of people. "Everyone of this group can do this". When you are referring to an undefined group of people of which any member can, you use "anyone". "Anyone can do this". I think.

pfac: The sentence does not define a group, the everyone is global. Anyone is less defined, true, but does not cover the global case imo.

PR #185, opened 2019-03-27
add team pictures

gabrielpoca:


src/components/home/Team/index.js
```diff
@@ -0,0 +1,75 @@
...
+import { StaticQuery, graphql } from "gatsby"
+
+import "./index.module.css"
+
+const socialLink = (platform, username) => {
+  switch (platform) {
+    case "twitter":
+      return `https://twitter.com/${username}`
+    default:
+      return "https://subvisual.co"
```
pfac: Why this default? Doesn't it make more sense to simply return `#`?

src/components/home/Team/index.js
```diff
@@ -0,0 +1,75 @@
+import _ from "lodash"
+import Img from "gatsby-image"
+import React from "react"
+
+import { StaticQuery, graphql } from "gatsby"
+
+import "./index.module.css"
+
+const socialLink = (platform, username) => {
```
pfac: Not that important, but `socialUrl` describes this function better, since we're not actually returning the link.

src/components/home/Team/index.js
```diff
@@ -0,0 +1,75 @@
...
+        <li styleName="member" key={name}>
+          <Img styleName="photo" fluid={photo.childImageSharp.fluid} />
+          <div styleName="name">
+            <span styleName="bold">{name}</span> <br />
+            {role}
+          </div>
+          <ul styleName="links">
+            {_.map(social, (name, platform) => (
+              <li styleName="link" key={platform}>
+                <a target="_blank" href={socialLink(platform, name)}>
```
pfac: Shouldn't this be using Gatsby `Link` component? Honest question, have no idea the tradeoff here, thought it was the convention.

gabrielpoca: That's just for internal navigation.

src/components/home/Team/index.js
gabrielpoca: Why? For accessibility it's better to have a valid link.

src/components/home/Team/index.js
gabrielpoca: I honestly don't think this matters. Many times we call "link" to "urls".

PR #186, opened 2019-03-28
accessible team section

gabrielpoca:


src/common/reset.css
```diff
@@ -32,3 +32,11 @@ ul {
 address {
   font-style: normal;
 }
+
+.visuallyhidden {
```
pfac: Shouldn't this be camelCase? Looks weird as a single word

src/components/home/Team/index.js
```diff
@@ -6,34 +6,64 @@ import { StaticQuery, graphql } from "gatsby"
...
+        linkName: "Github",
+        linkShortname: "Git",
+        url: `https://github.com/${username}`,
+      }
     default:
-      return "https://subvisual.co"
+      return {
+        linkName: "Other",
+        linkShortname: "Other",
+        url: "https://subvisual.co",
```
pfac: I still think this is not a great default value.

src/components/home/Team/index.js
```diff
@@ -6,34 +6,64 @@ import { StaticQuery, graphql } from "gatsby"
...
-          <ul styleName="links">
-            {_.map(social, (name, platform) => (
-              <li styleName="link" key={platform}>
-                <a target="_blank" href={socialLink(platform, name)}>
-                  link
-                </a>
-              </li>
-            ))}
-          </ul>
+          <nav aria-label="Social Links">
```
pfac: Why did you go with a `<nav>` here?

Looking at http://html5doctor.com/nav-element/, my intuition is that the `<nav>` tag should be used for the nav bar in the hero, but not for the social links of each team member, as it is not part of the navigation required to go around the website.

gabrielpoca: You're right 👍

src/components/home/Team/index.js
gabrielpoca: Why?

src/common/reset.css
gabrielpoca: 👍

src/components/home/Team/index.js
pfac: 1. Because `#` won't even cause a reload;
2. If you forgot to add the information for any social network, this won't warn you of anything, you'll have to get it by looking at the final result, where an error would warn a developer right away;
3. While this should redirect to the same page, it is an assumption not being made anywhere else;
4. We'll change to subvisual.com as our main domain with this rebranding.

PR #196, opened 2019-04-12
Add noscript fallbacks to Title and LoadPlaceholder

jfranciscosousa:
* Render the default title font while scripts are not loaded on the `Title` component
* Add noscript tag to the `LoadPlaceholder` component, rendering it's children without the loading animation if there are no scripts
* Extends `html.js` and removes the default `noscript` message (this app works well with Javascript)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "dependencies": {
     "classnames": "^2.2.6",
     "gatsby": "^2.1.18",
-    "gatsby-image": "^2.0.35",
+    "gatsby-image": "^2.0.38",
     "gatsby-plugin-google-analytics": "^2.0.18",
     "gatsby-plugin-manifest": "^2.0.19",
     "gatsby-plugin-offline": "^2.0.24",
--- a/src/components/LoadPlaceholder/index.js
+++ b/src/components/LoadPlaceholder/index.js
@@ -1,32 +1,45 @@
-import React, { Component } from "react"
+import React, { useState } from "react"
+import { renderToStaticMarkup } from "react-dom/server"
 import Observer from "react-intersection-observer"
+
+import useDetectJavascript from "../../common/useDetectJavascript"
 
 import "./index.module.css"
 
-export default class LoadPlaceholder extends Component {
-  state = { loaded: false, visible: false }
+export default function LoadPlaceholder({ delay = 0, dark, children }) {
+  const [loaded, setLoaded] = useState("")
+  const [visible, setVisible] = useState(false)
+  const hasJavascript = useDetectJavascript()
+  const loadedStyle = loaded ? "loaded" : ""
+  const darkStyle = dark ? "dark" : ""
 
-  onLoad = () => this.setState({ loaded: true })
+  const onLoad = () => setLoaded(true)
 
-  handleChange = isIntersecting => {
+  const handleChange = isIntersecting => {
     if (!isIntersecting) return
 
-    return this.setState({ visible: true })
+    setVisible(true)
   }
 
-  render() {
-    const delay = this.props.delay || 0
-    const dark = this.props.dark ? "dark" : ""
-    const loaded = this.state.loaded ? "loaded" : ""
-
-    return (
-      <Observer onChange={this.handleChange} styleName="root">
-        <div
-          style={{ transitionDelay: `${delay}s` }}
-          styleName={`placeholder ${dark} ${this.state.visible ? loaded : ""}`}
-        />
-        {this.props.children(this.onLoad)}
-      </Observer>
-    )
-  }
+  return (
+    <>
+      <noscript
+        dangerouslySetInnerHTML={{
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
   "dependencies": {
     "classnames": "^2.2.6",
     "gatsby": "^2.1.18",
-    "gatsby-image": "^2.0.35",
+    "gatsby-image": "^2.0.38",
```
pfac: Why the update?

src/components/LoadPlaceholder/index.js
```diff
@@ -1,32 +1,45 @@
...
-        />
-        {this.props.children(this.onLoad)}
-      </Observer>
-    )
-  }
+  return (
+    <>
+      <noscript
+        dangerouslySetInnerHTML={{
+          __html: renderToStaticMarkup(children(onLoad)).replace(
```
pfac: Why the replace?

src/components/PageTitle/index.module.css
```diff
@@ -6,6 +6,10 @@
   color: #045cfc;
 }
 
+.regular-font {
```
pfac: May I suggest `.with-tittle` instead? It explains the difference between this and the default in `.root`.

src/components/home/Hero/Title.js
```diff
@@ -3,21 +3,35 @@ import PropTypes from "prop-types"
...
-        <Planet morph={planetMorph} codeName="heroTittle" color="blue" />
-      </span>
-    </span>{" "}
-    <span styleName="glue">that empower</span> people
-  </PageTitle>
-)
+const Title = ({ planetMorph }) => {
+  const hasJavascript = useDetectJavascript()
+
+  if (hasJavascript) {
```
pfac: Can you change this to a guard clause? Less braces involved :stuck_out_tongue:

src/components/LoadPlaceholder/index.js
joaojusto: @pfac specific tricks to render under browsers that don't support `.js`

joaojusto: React does not support nested <noscript/> tags, this is a workaround for it.

package.json
joaojusto: Why not?

joaojusto:
@pfac the purposose of the html, as said on the description, is to remove the default noscript message.

pfac:
> @pfac the purposose of the html, as said on the description, is to remove the default noscript message.

Missed that. Carry on.

package.json
pfac: Are we updating stuff just because now?

I have nothing against it tbh, just wanted to know if there was any issue this solved. Carry on

src/components/LoadPlaceholder/index.js
jfranciscosousa: Already explained :slightly_smiling_face:

jfranciscosousa: But basically React itself doesn't handle nested `noscript` tags that well. And that `children` prop can have the default gatsby-image component, which already has a noscript tag.

src/components/PageTitle/index.module.css
jfranciscosousa: Sure

package.json
jfranciscosousa: During my adventure with the nested `noscript` issue I tried to update this just to see if something would be magically fixed. This was useless however. Should I revert it?

src/components/home/Hero/Title.js
jfranciscosousa: Of course :smile:

package.json
pfac: Nope, it's good, just wanted to know why.

src/components/home/Hero/Title.js
jfranciscosousa: But wait, this is already a guard clause no? I have no second `else`, I only use braces here because of the AirBnB style guide, which enforces braces on multi-line conditionals (I am used to that because we use it on every JS project). 

I can remove them if you really want to though :stuck_out_tongue:

src/components/PageTitle/index.module.css
jfranciscosousa: Will change it on the component props also

pfac: It's `tittle`, with 2 `t`s, it's the name of the dot in the i. :pray:

jfranciscosousa: 🙃 my bad

src/components/home/Hero/Title.js
jfranciscosousa: Oh yeah, but we still need to return some JSX in both cases, but we can "exit early" with the bad clause, that is when we don't JS available. I'll change it

PR #197, opened 2019-04-12
Add ParallaxBanners to Hero images

jfranciscosousa:
Uses `react-scroll-parallax` for the effect.


src/components/layout/index.js
```diff
@@ -7,16 +7,18 @@ import Footer from "./Footer"
...
-  <div styleName="root">
-    <Helmet>
-      <link rel="stylesheet" href="https://use.typekit.net/bcx8qfd.css" />
-    </Helmet>
-    <Header />
-    <main>{children}</main>
-    <Footer />
-  </div>
-)
+const Layout = ({ children }) => {
```
pfac: Why this change?

src/common/reset.css
```diff
@@ -36,8 +36,17 @@ address {
...
   position: absolute !important;
-  height: 1px; width: 1px;
+  height: 1px;
+  width: 1px;
   overflow: hidden;
   clip: rect(1px 1px 1px 1px); /* IE6, IE7 */
   clip: rect(1px, 1px, 1px, 1px);
 }
+
+.Parallax-outer-override {
```
pfac: Since these classes refer to a specific package, can you please put them in `src/common/react-scroll-parallax.css` and include it in `normalize.css`?

jfranciscosousa: Sure, good idea.

src/components/layout/index.js
jfranciscosousa: Nevermind, these are leftovers from some experiences. I'll revert this

jfranciscosousa:
I did not optimize the images, I'll do it. Didn't think about the repo size sorry :stuck_out_tongue:

joaojusto:
Only Pedro does :D

PR #201, opened 2019-06-06
Upgrade Gatsby to 2.8

pfac:
Why:

* Apparently there's a security vulnerability on `js-yaml` prior to
  version `3.13.1` (see https://github.com/nodeca/js-yaml/pull/480 for
  further details);
* Upgrading `js-yaml` requires upgrading `gatsby`;
* As a general rule, it's not a good idea to linger very far behind the
  base framework.

* Upgrading Gatsby and all Gatsby plugins, transformers, and directly
  related dependencies to the latest compatible versions.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10,17 +10,17 @@
     "@babel/highlight" "^7.0.0"
 
 "@babel/core@^7.0.0":
-  version "7.4.3"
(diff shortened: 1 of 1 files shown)


joaojusto:
@pfac sorry :/

PR #205, opened 2019-06-07
Upgrade react-intersection-observer

pfac:
Why:

* It is now above 1.0 (which means we can use semver to avoid breaking
  stuff).

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "version": "0.1.0",
   "author": "Kyle Mathews <mathews.kyle@gmail.com>",
   "dependencies": {
-    "@researchgate/react-intersection-observer": "^0.7.4",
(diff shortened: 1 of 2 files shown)


naps62:
@pfac I'm confused. why does 1.0 mean that? Can't you trust semver just the same in 0.x.x?

pfac:
> @pfac I'm confused. why does 1.0 mean that? Can't you trust semver just the same in 0.x.x?

@naps62 nope. `0.x` versions are for initial development, should be considered unstable and prone to change at any time. From the [Semantic Versioning Specification](https://semver.org/):

> 4. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

naps62:
uh. TIL

PR #203, opened 2019-06-07
Add splash screen

joaojusto:

PR #213, opened 2019-11-08
Add splash screen animation

jfranciscosousa:
Supersedes #203 

Use `react-morph` to morph the splash screen into the planet (tittle).
It also adds font subsetting to speed up the loading of the font used on the title.


pfac:
Please update the Node version in a separate PR.

src/components/splash_screen.module.css
```diff
@@ -0,0 +1,70 @@
...
+    height: 150vh;
+  }
+}
+
+@media only screen and (orientation: landscape) {
+  .splash {
+    width: 150vw;
+    height: 150vw;
+  }
+}
```
pfac: What's the idea here? Wouldn't `width: 100vw` and `height: 100vh` work for every case?

src/components/planet.js
```diff
@@ -1,61 +1,74 @@
...
   hoveringMax,
-  morph,
+  color,
 }) => {
   const hoveringAmplitude = _random(hoveringMin, hoveringMax)
   const hoveringAnimationDuration = _round(_random(0.8, 1.8), 2)
-  const animationName = `planet-${codeName}-hovering`
-  const radialID = `planet-${codeName}-radial`
+  const animationName = `planet-${codeName}-hovering-${_uniqueId()}`
+  const radialID = `planet-${codeName}-radial-${_uniqueId()}`
```
pfac: Why the unique IDs? Do we have repeating codenames?

src/components/splash_screen.module.css
jfranciscosousa: This was Justo's code. Let me double check and I'll get back to you.

src/common/normalize.css
```diff
@@ -1,6 +1,10 @@
-@import './fonts.css';
-@import './reset.css';
-@import './react-scroll-parallax.css'
+@import "./fonts.css";
+@import "./reset.css";
+@import "./react-scroll-parallax.css";
```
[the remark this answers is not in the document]
jfranciscosousa: My editor thought that was best. I'll revert.

src/components/home/hero_section/title.js
```diff
@@ -1,37 +1,85 @@
...
-  })
 
-  if (!hasJavascript)
+  function renderAnimatedLetters(string) {
+    return Array.from(string).map((letter, index) => (
+      <motion.span key={index} variants={letterVariants}>
+        {letter}
+      </motion.span>
+    ))
+  }
```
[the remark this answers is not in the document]
jfranciscosousa: I usually prefer writing functions explcitily at the root of files, but I can change that if you want.

src/components/planet.js
jfranciscosousa: Was trying to fix the Safari bug.

jfranciscosousa: Also, I'm keen on using that to make sure we have no conflicts. It's a pattern I used before when dealing with `id` tags. What do you think?

The problem on safari was because the ID was repeated when the `morph` cloned the elements. As you can notice I moved that out of the `{...moprh}` element.

I think this can prevent future bugs.

src/components/home/hero_section/title.js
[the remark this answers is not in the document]
pfac: I don't mind it, it was just a suggestion for cleaner code.

src/components/planet.js
pfac: > Was trying to fix the Safari bug.

Did you succeed?

> Also, I'm keen on using that to make sure we have no conflicts. It's a pattern I used before when dealing with id tags. What do you think?
> 
> The problem on safari was because the ID was repeated when the morph cloned the elements. As you can notice I moved that out of the {...moprh} element.
> 
> I think this can prevent future bugs.

I agree with the pattern, I've used it before and I second it, but in this specific case if you look through the codebase, the `codename` is unique to every instance of the planet precisely to avoid clashings with the animations and gradients. There is no need for the unique ID here because of that.

I'm not exactly aware anymore why, but I believe there was a reason for the `codename` to be set from above, it had to be used elsewhere. I fear this unique ID might be breaking something but I no longer remember exactly what.

However if we manage to remove the `codename` in the future, I would very much prefer to change it to use this approach.

jfranciscosousa: I'll remove the `uniqueId` and I'll get back to you, to check if it's still working on Safari.

src/components/splash_screen.module.css
jfranciscosousa: Yeah, this has some logic to it. Basically this sets the splash screen to be much larger than the actual screen, otherwise, when morphing it would look pretty strange. The difference between the `vw` and `vh` is to basically use the largest unit, on portrait `height > width` so we use height. Same stuff for landscape.

jfranciscosousa: This way, when we start the morph, the planet would be bigger than screen giving us a smoother effect.

PR #214, opened 2019-11-13
Fix logo alignment on the about section

jfranciscosousa:
The logo was not aligned on any browser that wasn't Firefox.


jfranciscosousa:
Tested on all breakpoints, seemed fine. But perhaps we can ask Baila about it

PR #215, opened 2019-11-15
Update Node version to the latest LTS

jfranciscosousa:
* Updates the Node version on the `.tool-version`
* Updates the engines field to force the node version


pfac:
Well [apparently](https://app.netlify.com/sites/subvisual/deploys/5dcefc632dd762000a7aa548) Netlify does not allow changing the NodeJS version. Seems we're stuck with `10.17.0` for the time being.

PR #220, opened 2019-11-22
Refactor intersection observer and increase view threshold

jfranciscosousa:
Why:
- The intersection observer mechanism was not optimal according to the designers.
- We needed a way to better control the threshold

How:
- Removing the previous intersection observer library and replace it with a more up to date one on `react-intersection-observers`, using hooks instead of render props.
- Removing uneeded components that were superseded by the hooks approach
- Removing the `LoadPlaceholder` from the team and footer section (will request validation from the design team).


pfac:
Relevant information: https://trello.com/c/IF4o3QEU/

jfranciscosousa:
@pfac probably we want to stall this PR so we can then apply any relevant linters

pfac:
Is this ready for review?

jfranciscosousa:
For review and merge yes

PR #223, opened 2020-01-22
reduce the website in 2mb on desktop

gabrielpoca:
In this change, I update the ventures' portfolio section to use a single
image for each venture. In the current implementation, we
load a color and a grey image for each venture. With this change, we
simply apply a css filter to grayscale the color image. This reduces the
size of the whole website in 2mb in desktop.

After this we can probably clean-up some css and assets, but I don't have time to do that now.


src/components/home/ventures_section/portfolio_subsection.js
```diff
@@ -8,9 +8,10 @@ import Venture from "./portfolio_subsection/venture"
 import styles from "./portfolio_subsection.module.css"
 
 const PortfolioSubsection = ({ planetMorph, ventures }) => {
+  const baseDelay = 0.3
```
pfac: Why this change? `baseDelay` is not being used outside `renderVenture` is it?

src/components/image_loader/index.module.css
```diff
@@ -0,0 +1,7 @@
+.root {
+  position: relative;
+
+  height: 100%;
+
+  overflow: hidden;
+}
```
pfac:
```suggestion
}

```

src/components/home/ventures_section/portfolio_subsection.js
gabrielpoca: I was going to move it outside the component. Just because it's a constant.

gabrielpoca:
@pfac pseudo selectors don't usually play well with css in js, but I'll try

gabrielpoca:
@pfac updated

PR #224, opened 2020-01-22
reduce the size from 15mb to 4.5mb in desktop

gabrielpoca:
Run imgoptim and reduce image quality overall.


pfac:
Shame on the deceased Justo for insisting there was no need to optimize the images with the argument that Gatsby would automatically take care of it. Thanks for proving me right :heart:

gabrielpoca:
He was sort of right xd gatsby does optimize, but you need to tell it how much xd

PR #225, opened 2020-01-30
stop loading both mobile and desktop images

gabrielpoca:
in some places, we use different images between mobile and desktop.
it's not just a matter of the images having a different size, they
also have different "cuts". because the images are loaded async with JavaScript, the CSS rules that add
'display: none' to mobile images in desktop (and the other way around)
are ignored. this means that desktop and mobile images are always loaded for every request.

to solve this, there's a thing called [art
direction](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Art_direction)
that [gatsby also supports](https://www.gatsbyjs.org/packages/gatsby-image/#art-directing-multiple-images).

extra: reduced the quality of hero-3 in desktop because it's the biggest
image in our website and it doesn't seem to change its final quality.

### DIFF
--- a/src/components/home/hero_section.js
+++ b/src/components/home/hero_section.js
@@ -27,7 +27,10 @@ const ParallaxImage = ({
                 fadeIn={false}
                 onLoad={onLoad}
                 fluid={[
-                  mobileImage,
+                  {
+                    ...mobileImage,
+                    media: "(max-width: 399px)",
+                  },
                   {
                     ...desktopImage,
                     media: "(min-width: 400px)",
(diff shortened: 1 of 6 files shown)


gabrielpoca:
@pfac I think we can merge this. The issues pointed out by Baila are not related to this PR and I'm going to fix them next.

src/components/home/hero_section.js
```diff
@@ -21,7 +26,13 @@ const renderParallaxImage = ({ baseDelay, image, parallaxAmount }) => (
               <Img
                 fadeIn={false}
                 onLoad={onLoad}
-                fluid={image}
+                fluid={[
+                  mobileImage,
```
pfac: Doesn't the mobile image need a limiting media rule like for team member images?

gabrielpoca:
@pfac updated for cross browser support

PR #227, opened 2020-01-31
Refactor meta information and add meta image

gabrielpoca:
* update the meta tags using with the recommendation in https://metatags.io
* change title (check the current title in the image bellow)
* make the code more explicit
* add a meta image so our website doesn't look like a dummy when shared on social media.

Screenshot 2020-01-31 at 16 45 04


src/pages/index.js
```diff
@@ -52,10 +52,7 @@ const IndexPage = () => {
         </style>
       </noscript>
 
-      <SEO
-        title="Subvisual - We nurture ideas that empower people"
```
pfac: If I didn't misunderstood any part here, this change is gonna make it so the home page has just `Subvisual` as the title. Is this intended?

As per SEO recommendations, we can and should use our _why_ here, since it is the closest we have to a slogan.

gabrielpoca: That's not what I read regarding SEO. And you can see that everywhere it shows the title "Subvisual - We nurture ideas that empower people" and then the description is exactly the same "We nurture ideas that empower people". So until we have proper copy for this, at least this remove the nonsense duplication.

pfac: It is not nonsense, that duplication is was intended. It is a common SEO recommendation that you include the keyphrase in both the title, and slug if applicable. Even if you actually repeat the phrase instead of using slightly different wording. See for instance https://yoast.com/wordpress/plugins/seo/titles-and-meta-descriptions/.

Despite the repetition, from the SEO perspective, it is highly recommended not to use only the brand name as the title. For instance, search for Zara, Thoughtbot, or Uphold. Some better than others but including a captivating sentence alongside the name is the recommendation.

gabrielpoca: There are two things to this:

1. keyword density is no longer a thing and it hasn't been a thing for a very long time. There's a lot of writing on the fact that it's myth.
2. "We", "nurture", "ideas", "that", "empower" and "people" and terrible keywords. There are way better ones in the body of the page, which is also indexed. Let's not try to shuffle some random keywords all over the place when they are not relevant.

PR #228, opened 2020-01-31
Improve the hero images

gabrielpoca:
* replace ParallaxBanner with Parallax to remove that weird zoom on the
  images.
* for some reason, there's a media query that removes grid-template-rows
  bellow 320px, which removes the images. I don't think this is intended
  behaviour.
* replace the LoadPlaceholder with ImageLoader component which is
  similar but much better.
* replace the desktop images with new ones exported from figma.
* add an alt to the images.

The change will turn this

Screenshot 2020-01-31 at 17 26 13

into this

Screenshot 2020-01-31 at 17 25 55


src/components/home/hero_section.js
```diff
@@ -1,52 +1,73 @@
...
+        overflow: "hidden",
+        position: "unset !important",
+      }}
+    >
+      <Parallax
+        y={[-15, 15]}
+        styleInner={absoluteStyle}
+        styleOuter={{
+          ...absoluteStyle,
+          marginTop: size.width > 400 ? desktopTopMarign : mobileTopMarign,
```
pfac: :warning: Mar**gin**

src/components/home/hero_section.js
```diff
@@ -1,52 +1,73 @@
...
-    ]}
-  />
-)
+  objectPosition,
+  mobileTopMarign,
+  mobileBottomMarign,
+  desktopTopMarign,
+  desktoBottomMarign,
+}) => {
+  const size = useWindowSize()
```
pfac: Why not destruct the variable to have `width`? I might be missing something but looks like it isn't even using the height.

src/utils/use_window_size.js
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect, useState } from "react"
+
+export default function useWindowSize() {
+  const isClient = typeof window === "object"
+
+  function getSize() {
```
pfac: Why not a simple guard returning `{}` if `isClient` is falsy?

src/components/home/hero_section.js
gabrielpoca: It's from https://usehooks.com/useWindowSize/. I would rather keep it as it is because it's general purpose and that way I can just copy it over again in case there's a bug and the code gets updated.

src/utils/use_window_size.js
gabrielpoca: save as above

PR #230, opened 2020-02-21
Change logo to assume color of text

pfac:
Why:

* Currently we're setting the color of the logo manually as a component
  property. However, when the background changes we usually want to
  change the color of the logo automatically. Instead of having logic to
  detect which color is being used in the background, we can instead
  change the logo to use the text color automatically.

* Refactoring the Logo component to always use `currentColor` (the
  current color of text);
* Refactoring the header and footer to set the text color accordingly.


src/components/logo.js
```diff
@@ -23,7 +19,7 @@ const Logo = ({ color }) => {
...
-          fill={colorCode}
+          fill="currentColor"
         />
         <defs>
           <radialGradient
```
jfranciscosousa: IE and Edge have a few quirks with `currentColor` on gradients. Can you perhaps test this there? Just to confirm.

pfac: It does not work at all in IE 11, but the website is completely screwed in there anyway, so I wouldn't bother with it.

It works just fine on Edge.

Tested with https://deploy-preview-230--subvisual.netlify.com/.

jfranciscosousa: Well, if IE 11 is not a thing, go for it :+1:

PR #231, opened 2020-02-21
Inject logo into header component

pfac:
Why:

* The blog will use a slightly different logo in the header, otherwise
  it should be the same header.

* Refactoring the layout header component to accept a function that
  renders the logo;
* Refactoring the layout component to accept a function that renders the
  header logo, falling back to the default logo component.

### DIFF
--- a/src/components/layout.js
+++ b/src/components/layout.js
@@ -2,17 +2,18 @@ import React from "react"
 import PropTypes from "prop-types"
 import Helmet from "react-helmet"
 
+import Logo from "./logo"
 import Header from "./layout/header"
 import Footer from "./layout/footer"
 
 import styles from "./layout.module.css"
 
-const Layout = ({ children }) => (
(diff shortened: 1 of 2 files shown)


src/components/layout.js
```diff
@@ -2,24 +2,30 @@ import React from "react"
...
-    <Header />
+    <Header renderLogo={renderHeaderLogo} />
     <main>{children}</main>
     <Footer />
   </div>
 )
 
 Layout.propTypes = {
   children: PropTypes.node.isRequired,
+  renderHeaderLogo: PropTypes.func,
```
jfranciscosousa: Should be required, otherwise :boom:

pfac: There's a default value for this property on line 28. How would this blow up?

jfranciscosousa: Completely drunk. You are right

PR #232, opened 2020-02-21
Change links to use current text color

pfac:
Why:

* Links are currently defaulting to the "black" color, and are changed
  using a component property. However their color is based on the
  background, which is set by parent components. So we can instead make
  links depend on text color, which is set by the same components that
  change the background.

* Refactoring the link component to always use the current color;
* Refactoring the about subsection component to set the current text
  color accordingly.

### DIFF
--- a/src/components/link.js
+++ b/src/components/link.js
@@ -5,8 +5,8 @@ import classNames from "classnames"
 
 import styles from "./link.module.css"
 
-const Link = ({ blank, children, color, faded, internal, size, title, to }) => {
-  const className = classNames(styles.root, styles[size], styles[color], {
+const Link = ({ blank, children, faded, internal, size, title, to }) => {
+  const className = classNames(styles.root, styles[size], {
     [styles.faded]: faded,
   })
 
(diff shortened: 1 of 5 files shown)


src/components/link.js
```diff
@@ -5,8 +5,8 @@ import classNames from "classnames"
 
 import styles from "./link.module.css"
 
-const Link = ({ blank, children, color, faded, internal, size, title, to }) => {
-  const className = classNames(styles.root, styles[size], styles[color], {
+const Link = ({ blank, children, faded, internal, size, title, to }) => {
```
jfranciscosousa: Why not keep the color prop, and use the `currentColor` as a default value? I think that would be best.

pfac: But why? As this PR shows, there are no instances where we need the link to be different than the current text color. If we need it we can always add it back, but for now there's no use for such a prop.

jfranciscosousa: Was a "just in case" scenario, but alas. Let it be

PR #236, opened 2020-02-22
Add support for inactive team members

pfac:
Why:

* The blog will need to fetch the information of users from the team
  data file, but this file is currently omitting past blog post authors.
  Ideally the author information should be in the post itself to keep
  things denormalized, but since it isn't, the quickest way to address
  the issue is to re-add old team members, and simply add a field to
  mark them as inactive, excluding them from the team page if so.

### DIFF
--- a/src/components/home/about_section/team_subsection.js
+++ b/src/components/home/about_section/team_subsection.js
@@ -7,7 +7,7 @@ import styles from "./team_subsection.module.css"
 
 const teamQuery = graphql`
   query TeamQuery {
-    allTeamYaml {
+    allTeamYaml(filter: { active: { ne: false } }) {
       edges {
         node {
           id
--- a/src/data/team.yaml
+++ b/src/data/team.yaml
@@ -1,4 +1,7 @@
 ---
+- name: Bruno Azevedo
+  active: false
+
 - name: "Fernando Mendes"
   photo:
     horizontal: "../images/team/fernando-mendes-h.jpg"


src/components/home/about_section/team_subsection.js
```diff
@@ -7,7 +7,7 @@ import styles from "./team_subsection.module.css"
 
 const teamQuery = graphql`
   query TeamQuery {
-    allTeamYaml {
+    allTeamYaml(filter: { active: { ne: false } }) {
```
jfranciscosousa: I think thaving `active == true` would be more readable no? Just my opinion anyway, feel free to merge 👍

pfac: If it works, I'm a bit noobish with GraphQL, I usually follow along GraphiQL. Lemme see

pfac: Apparently that's not a thing for Gatsby :man_shrugging:

jfranciscosousa: `filter: { active: { eq: true } }` did not work? Weird

pfac: Oh, I thought you actually were saying to use a different syntax like `filter: { active == true }`.

`filter: { active: { eq: true } }` works but doesn't do what we want here. In this specific case, and to avoid adding unnecessary fields to the data file, we want to exclude all those team members for which the `active` field is explicitly `false`. If unset, we'll default to assuming the team member is active. If we ask GraphQL for the team members that have the `active` field with `true`, none would be shown.

jfranciscosousa: Got it :+1:

PR #239, opened 2020-03-10
Set base typography sizes

pfac:
Why:

* The links in the top navigation bar are too small;
* Most of the content of our pages use a base font size and line heights
  for the body of text, and we're gradually changing components that
  simply inherit these values to do just that. The top navigation bar
  was already changed accordingly.

### DIFF
--- a/src/common/normalize.css
+++ b/src/common/normalize.css
@@ -4,12 +4,18 @@
 
 body {
   overflow-x: hidden;
-}
 
-html,
-body {
   color: #403f4c;
 
   font-family: colfax-web, sans-serif;
+  font-size: 16px;
   font-weight: 400;
+  line-height: 20px;
+}
+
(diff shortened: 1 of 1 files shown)


src/common/normalize.css
```diff
@@ -4,12 +4,18 @@
...
 body {
   overflow-x: hidden;
-}
 
-html,
-body {
   color: #403f4c;
 
   font-family: colfax-web, sans-serif;
+  font-size: 16px;
```
jfranciscosousa: Are these values on variables somewhere? Just a question

pfac: Nope, this project does not use any variables whatsoever, both for flexibility and speed at the time this design was implemented.

jfranciscosousa: Yes 👍

PR #243, opened 2020-03-16
Refactor splash screen component

pfac:
Why:

* The splash screen is, by definition, shown only for a brief amount of
  time; however it is the surrounding page that's handling the logic of
  hiding it after a given period (or when Javascript is disabled),
  making the page code more complex than it should be;
* The SVG code of the logo in the splash screen does not seem to have
  been optimized, and is using hard-coded IDs which may cause hard to
  debug conflicts;
* The splash screen logo is showing up with a different size on Firefox
  and Chrome, both smaller than it should be.

This change addresses the issues by:

* Moving the hiding logic to the splash screen component, instead using
  a handler for the parent to hook into and react to a change;
* Optimizing the SVG code for the logo screen and refactoring it to use
  dynamic IDs;
* Changing the splash screen styles to treat the logo as a block, and to
  stretch its container to the full width of the parent.


src/components/splash_screen.js
```diff
@@ -1,44 +1,84 @@
-import React from "react"
+import React, { useEffect, useState } from "react"
+import PropTypes from "prop-types"
+import { disablePageScroll, enablePageScroll } from "scroll-lock"
+import uuid from "uuid/v4"
```
jfranciscosousa: I don't think you should import this given that we already have `lodash/uniqueId` and it performs the same function and we don't need to fill in extra bundle size :+1:

src/pages/index.js
```diff
@@ -31,33 +30,23 @@ const config = { spring, easings, getMargins: true }
...
-      <Splash show={renderSplash} morph={morph} />
-      <HeroSection planetMorph={morph} hidePlanet={renderSplash} />
-      <VenturesSection />
-      <AboutSection />
-    </Layout>
+    <>
+      <SplashScreen
+        lockScrollFor={1000}
+        morph={morph}
+        onHide={() => setHidingTittlePlanet(false)}
```
jfranciscosousa: Advise against inline arrow functions, regardless of what Gabriel might say :joy:

jfranciscosousa:
Overall the website is showing before the splash screen is applied :thinking: looking at the code I can't find the problem.

src/components/splash_screen.js
pfac: Nice, didn't know about `_.uniqueId`. Seems like I have a new optimization to do :stuck_out_tongue:

src/pages/index.js
pfac: Do you have any reason for that advice or is it subjective?

Both ways work for me actually. Using a variable allows the component usage to be more semantic and readable. On the other hand functions simple as this one are fairly easy to read and the code remains simpler and more succint.

Changed anyhow :+1:

pfac:
@jfranciscosousa it seems the issue has been here all along, but was hidden by the root container having the same background as `.splash`. The splash is always hidden at first, but the root container having the exact same background gives the illusion it was always there.

Turns out your workaround is actually the proper way to get it to work, at least until we can prevent react-morph from hiding our stuff.

src/pages/index.js
jfranciscosousa: Basically it's changing the reference every time the parent re-renders. If the parent re-renders that inline expression is evaluated and changes it's reference, causing a re-render of the `SplashScreen` component, even if the other props don't change. By defining the function early, you maintain that reference and avoid extra renders.

jfranciscosousa: But performance impacts are negligible, usually, the Pong game we made with React would suffer tremendously with this but we ported everything to `requestAnimationFrame`  anyhow so there were almost no React re-renders.

But basically it's your first option. I like defining variables with functions, makes stuff more readable. I write components usually with `hooks - variables - handlers - jsx`. So people always expect extra behavior in the first section of the component, and the `jsx` is just the structure.

jfranciscosousa:
The logo on mobile is a bit bigger, fix that and it's :green_apple:

PR #252, opened 2020-03-20
Add blog contributors data

pfac:
Why:

* Some blog post authors are no longer on the team, some were just
  guests.


pfac:
Forever in our :heart:

PR #257, opened 2020-03-20
Add blog post body component

pfac:
Why:

* The body of a blog post will be generated from the respective Markdown
  file. As such, we need to style the generated elements without being
  able to set specific classes to each one.
* There are also many details here that have to tweaked all at once
  since we can't split the inner parts of the Markdown document into
  smaller components, such as spacings, typography, adjustments for
  responsiveness, etc.

### DIFF
--- a/src/components/blog/post/body.js
+++ b/src/components/blog/post/body.js
@@ -7,9 +7,8 @@ import styles from "./body.module.scss"
 
 const PostBody = ({ html }) => (
   <BodyWrapper>
-    {/* eslint-disable react/no-danger */}
+    {/* eslint-disable-next-line react/no-danger */}
     <div className={styles.root} dangerouslySetInnerHTML={{ __html: html }} />
-    {/* eslint-enable react/no-danger */}
   </BodyWrapper>
 )
 
--- a/src/components/blog/post/body.js
+++ b/src/components/blog/post/body.js
@@ -1,0 +1,20 @@
+import React from "react"
+import PropTypes from "prop-types"
+
+import BodyWrapper from "./body_wrapper"
+
+import styles from "./body.module.scss"
+
+const PostBody = ({ html }) => (
+  <BodyWrapper>
+    {/* eslint-disable react/no-danger */}
+    <div className={styles.root} dangerouslySetInnerHTML={{ __html: html }} />
+    {/* eslint-enable react/no-danger */}
+  </BodyWrapper>
+)
+
+PostBody.propTypes = {
+  html: PropTypes.string.isRequired,
+}
+
+export default PostBody
--- a/src/components/blog/post/body.module.scss
+++ b/src/components/blog/post/body.module.scss
@@ -1,14 +1,4 @@
 .root {
-  font-family: "Acta", serif;
-  font-size: 16px;
-  font-weight: 400;
-  line-height: 28px;
-
-  @media (--desktop) {
-    font-size: 20px;
-    line-height: 32px;
-  }
-
   @import './body/anchor';
   @import './body/blockquote';
   @import './body/bold';
@@ -20,6 +10,16 @@
   @import './body/italic';
   @import './body/list';
   @import './body/paragraph';
+
+  font-family: "Acta", serif;
+  font-size: 16px;
+  font-weight: 400;
+  line-height: 28px;
+
+  @media (--desktop) {
+    font-size: 20px;
+    line-height: 32px;
+  }
 
   /* Drop cap */
 
--- a/src/components/blog/post/body.module.scss
+++ b/src/components/blog/post/body.module.scss
@@ -1,0 +1,46 @@
+.root {
+  font-family: "Acta", serif;
+  font-size: 16px;
+  font-weight: 400;
+  line-height: 28px;
+
+  @media (--desktop) {
+    font-size: 20px;
+    line-height: 32px;
+  }
+
+  @import './body/anchor';
+  @import './body/blockquote';
+  @import './body/bold';
+  @import './body/code';
+  @import './body/code_block';
+  @import './body/heading';
+  @import './body/horizontal_rule';
+  @import './body/image';
+  @import './body/italic';
+  @import './body/list';
+  @import './body/paragraph';
+
+  /* Drop cap */
+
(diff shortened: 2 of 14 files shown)


pfac:
@jfranciscosousa in terms of code review, this is the big one. Any tips to improve the readability or organization of these styles is more than welcome. :pray: 

(I'm actually considering documenting the hell out of these classes)

src/components/blog/post/body.module.css
```diff
@@ -0,0 +1,358 @@
+.root {
+  font-family: "Acta", serif;
+  font-size: 16px;
+  font-weight: 400;
+  line-height: 28px;
+}
+
+/* Limit block elements w/ paddings */
+
+.root blockquote,
```
jfranciscosousa: Don't we have `sass` like nesting here? I think that would be prettier and easy to read.

src/components/blog/post/body.module.css
```diff
@@ -0,0 +1,358 @@
...
+  margin-top: 20px;
+}
+
+.root p:not(:last-child) {
+  margin-bottom: 28px;
+}
+
+/* Images (in paragraphs) */
+
+.root p > img {
```
jfranciscosousa: If you want, there is a plugin that unwraps images out of the `p` tags that the `remark` engine creates. I think accesibility wise that is a bit stupid, to have an image inside a paragraph. https://www.gatsbyjs.org/packages/gatsby-remark-unwrap-images/

src/components/blog/post/body.module.css
```diff
@@ -0,0 +1,358 @@
...
+  width: 25vw;
+  max-width: 174px;
+  margin: 28px auto;
+
+  border-top: none;
+  border-bottom: currentColor solid 1px;
+}
+
+@media (min-width: 619px) {
+  /* Images (in paragraphs)
```
jfranciscosousa: :rotating_light:

src/components/blog/post/body.module.css
pfac: Is that a thing with CSS modules? :open_mouth: 

Just tested and seems it ain't. But if that's a possibility without adding junk I want it!

src/components/blog/post/body.module.css
pfac: So, no comments is better? I'm honestly looking for any help making this file not look like a mess.

src/components/blog/post/body.module.css
pfac: Adding `postcss-nested` in #260 :heart:

src/components/blog/post/body.module.css
jfranciscosousa: No :joy: the formatting was weird, doesn't prettier fix that?

pfac: On my setup, prettier does not touch what is not meant to be touched. And in our case only stylelint touches CSS files :stuck_out_tongue: 

Anyhow this is going away now :grin:

pfac:
@jfranciscosousa updated with all the macumbas from #262, and #266.

src/components/blog/post/body.js
```diff
@@ -0,0 +1,20 @@
+import React from "react"
+import PropTypes from "prop-types"
+
+import BodyWrapper from "./body_wrapper"
+
+import styles from "./body.module.scss"
+
+const PostBody = ({ html }) => (
+  <BodyWrapper>
+    {/* eslint-disable react/no-danger */}
```
jfranciscosousa: `// eslint-disable-next-line react/no-danger` :upside_down_face:

src/components/blog/post/body.module.scss
```diff
@@ -0,0 +1,46 @@
...
+  font-size: 16px;
+  font-weight: 400;
+  line-height: 28px;
+
+  @media (--desktop) {
+    font-size: 20px;
+    line-height: 32px;
+  }
+
+  @import './body/anchor';
```
jfranciscosousa: Probably we could put the imports all at the top of the rule, no?

src/components/blog/post/body/_anchor.scss
```diff
@@ -0,0 +1,17 @@
...
+    color: #2421ab;
+
+    font-weight: 700;
+    text-decoration: none;
+  }
+
+  &:hover {
+    opacity: 0.5;
+
+    transition: opacity 0.2s;
```
jfranciscosousa: Is the designer happy with the default easing? :joy:

src/components/blog/post/body/_heading.scss
```diff
@@ -0,0 +1,54 @@
+h1 {
```
jfranciscosousa: ???

src/components/blog/post/body.js
pfac: Nice one, changed in 93dfdaf.

src/components/blog/post/body.module.scss
pfac: Sure, changed in 93dfdaf.

src/components/blog/post/body/_anchor.scss
pfac: :man_shrugging: Guess so, this was copied from the Link component (https://github.com/subvisual/subvisual.com/blob/master/src/components/link.module.css) which he never complained about.

src/components/blog/post/body/_heading.scss
pfac: This is an old fix in the Subvisual blog. Funny story, we used to allow users to use any headers they wanted, which meant many blog posts used multiple `h1` tags.

For SEO reasons, in most cases a page should not have more than one `h1` tag. But, since we're dealing with Markdown here, there's nothing stopping everyone from putting their own H1 in there, although we already have it in the blog post header (Markdown only generates the body).

This ensures such _meliantes_ don't screw up their own blog posts by mistake.

jfranciscosousa: So if someone by mistake adds an `#` thing on the mardown, that gets hidden right?

pfac: Precisely.

jfranciscosousa: Makes sense

PR #260, opened 2020-03-20
Add PostCSS plugin for nesting CSS

pfac:
Why:

* The styles for the blog post body (being added in #257) are more
  complex than resonable mostly because, due to the Markdown rendering,
  we're unable to use specific CSS classes for each rendered element,
  and as such have to rely heavily on descendant selectors, which is
  causing a lot of repetition.
* The issue is not specific to the blog though, and as shown with the
  CTA component, this can significantly simplify the styles of some
  components.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -1,3 +1,6 @@
+const postCssCustomMedia = require(`postcss-custom-media`)
+const postCssNested = require(`postcss-nested`)
+
 module.exports = {
   siteMetadata: {
     title: "Subvisual",
@@ -10,7 +13,7 @@ module.exports = {
     {
       resolve: `gatsby-plugin-postcss`,
       options: {
-        postCssPlugins: [require(`postcss-custom-media`)()],
+        postCssPlugins: [postCssCustomMedia(), postCssNested()],
       },
     },
     `gatsby-plugin-react-helmet`,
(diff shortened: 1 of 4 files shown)


gatsby-config.js
```diff
@@ -10,7 +13,7 @@ module.exports = {
     {
       resolve: `gatsby-plugin-postcss`,
       options: {
-        postCssPlugins: [require(`postcss-custom-media`)()],
+        postCssPlugins: [postCssCustomMedia(), postCssNested()],
```
jfranciscosousa: This works fine, but you can check out `precss` for future times. It's what I use most times. It emulates `SASS` pretty well. You can check out the plugins there.

`autoprefixer` would also work pretty well here.

PostCSS is literally the best thing basically.

There is `cssnano` to minify and then `postcss-purgecss` to remove any CSS you don't use if you import something like Tailwind or Bulma, to remove stuff you are not using. Kinda like tree shaking

pfac: Like. `precss` actually actually handles nesting through `postcss-nested`, that's why I went with it. I just refrained from using the entire package because if we're not using it there's no reason I guess. :man_shrugging: 

Have to look into the others. Thanks :heart:

jfranciscosousa: Perfectly reasonable :+1:

PR #262, opened 2020-03-27
Replaces postcss-nested with Gatsby SASS

pfac:
Why:

* Trying to cleanup the post body styles in #212, the best approach I
  found was to use SASS to import styles directly in the middle of other
  files. I looked into `precss`, but turns out there is a recommended
  Gatsby plugin that uses Node SASS underneath, giving us a more battle
  tested support for SASS. I also tried `precss` and wasn't able to
  achieve quite what I had in mind, but it worked with the Gatsby SASS
  plugin. :shrug:


src/components/call_to_action.js
```diff
@@ -1,7 +1,7 @@
 import React from "react"
 import PropTypes from "prop-types"
 
-import styles from "./call_to_action.module.css"
+import styles from "./call_to_action.module.scss"
```
jfranciscosousa: Not working with `precss`? What was the problem?

jfranciscosousa: Or better, what were you trying to do exactly? Nothing wrong with using `node-sass`. I just avoid it because sometimes it just explodes on my face (binary compilations and whatnot)

pfac: It has been a few experiments since then, so I'm not 100% sure by now, but you can see the end result I was looking for here: https://github.com/subvisual/subvisual.com/pull/212/files#diff-af53ed79ac06ecc26484e4bef623f8b3R14.

If I'm not mistaken, `precss` was processing those imported files on their own, without nesting them where I was importing them, which obviously didn't generate the styles we needed.

jfranciscosousa: Because you are importing them inside a CSS rule? :joy: You should do that with mixins though

jfranciscosousa: And you can keep out the node-sass :tractor:

jfranciscosousa: With mixins, you can just export the internal rules (tag agnostic)
```
@mixin blockquote {
  some styles

  &:hover {
    bla bla
  }
}

then on the blockquote

blockquote {
  @include blockquote;
}
```

jfranciscosousa: Notice that you should keep the mixins tag agnostic, so you can even re-use them in other kinds of tags like

```
h1, h2, h3, h4, h5 {
  @include headerBase;
}
```

pfac: That's precisely why I didn't go with mixins. :wink:

Since these are not reusable tag agnostic styles, mixins would be a poor choice in my opinion. This approach of importing the styles in place has many advantages, while I do agree it is unorthodox:
- The styles in these files reflect the point of view of the HTML generated from the Markdown files, where there are only tags, no classes.
- These files are also partials, their name suggests they're not meant to be used as standalone stylesheets.
- They are also inside a director named after the only component that's supposed to use them.
- And finally, the fact we can inline them in the body component CSS module both keeps things readable and avoids exceeding the maximum 3 levels of nesting our linter allows.

I understand the feeling regarding `node-sass` here, but I believe these reasons make this a option far better than the mixins. :grimacing: 

Do you have any other suggestions?

jfranciscosousa: Cool, but what about just importing them at root level? I think `postcss` just can't import stuff inside a block

jfranciscosousa: Forget it, wouldn't work them. If you think it's best, go ahead. `node-sass` it's not too bad, until it crashes on netlify :joy:

pfac: That breaks the first of my points. You end up with a partial looking like this:

```scss
.root p {
  /* whatever makes a p a p */
}
```

Looking at such a file one would invariably wonder what the hell is that `.root` class. Markdown generates no such thing. Or does it? Best not to confuse everybody. :stuck_out_tongue:

PR #273, opened 2020-04-01
Dynamically fill blog posts list

pfac:
Why:

* Last step before #212, we need to change https://subvisual.com/blog to
  dynamically retrieve the list of posts for which we generated a page.

* Changing the posts list component to query the contents processed from
  Markdown files;
* Adding a formatted date field to the posts list entry component, along
  with the intro in the frontmatter, and the author information.

### DIFF
--- a/src/components/blog/posts_list.js
+++ b/src/components/blog/posts_list.js
@@ -43,7 +43,7 @@ const renderItem = ({ fields, frontmatter }) => {
   )
 }
 
-const BlogPostList = () => {
+const BlogPostsList = () => {
   const {
     allMarkdownRemark: { nodes: postsNodes },
   } = useStaticQuery(query)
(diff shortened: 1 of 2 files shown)


src/components/blog/posts_list.js
```diff
@@ -1,23 +1,71 @@
...
+          title
+          intro
+        }
+      }
+    }
+  }
+`
+
+export default () => (
+  <StaticQuery
```
jfranciscosousa, line 49: You can use `useStaticQuery` and not use the horrendous pattern that a render prop is

pfac: I love you right now :heart:

jfranciscosousa: My freelancing sites all use a bunch of hooks sprayed on a `data` folder. 

Images, copy in markdown, everything is there. My components just import those and magically there are images everywhere and pretty copy everywhere. If you know what you are doing Gatsby is bar to none <3

PR #275, opened 2020-04-02
Upgrade stylelint

pfac:


jfranciscosousa:
You should apply the fix that we made on the Filecoin project. Actually use the `stylelint-prettier-plugin`

pfac:
> You should apply the fix that we made on the Filecoin project. Actually use the `stylelint-prettier-plugin`

Plugin added. Thanks :+1:

PR #286, opened 2020-04-03
Add links in blog post body in a new tab

pfac:


pfac:
@naps62 I believe this is what you were looking for.

PR #288, opened 2020-04-04
Fix blog post header margins

pfac:
Why:

* The desktop bottom margin is not being applied in production;
* There should not be any side margin, the placement is handled by the
  inner wrapper components.


pfac:
Any production only process that we might be using is coming from Gatsby itself, nothing I'm aware we did ourselves :man_shrugging:

PR #290, opened 2020-04-05
Refactor hero title component

pfac:
Why:

* The logic to animate the hero title is mixed with the rest of the
  component, which makes it next to impossible to disable animations.
* When the hero tittle is the planet hovering, we're using a different
  unicode character to print an `i` without the tittle. However this
  messes the content. For instance, if you copy and paste the title, the
  character copied is not an `i`.
* Despite using a different character, we're using a different font
  family for the tittleless `i`, which is Base64 encoded in the fonts
  stylesheet.

This change addresses these issues by:

* Extracting the animation logic in the hero section title component
  into its own sub-component.
* Refactoring the content of the title to be properly copyable, and use
  a proper `i`.
* Refactoring the planet component to remove the `hide` property,
  instead the component using it decides whether or not to render it.
* Replacing the Base64 encoded subset font with one containing only the
  `i` but without the tittle in the glyph.


src/components/home/hero_section/title/animated_letters.js
```diff
@@ -0,0 +1,89 @@
...
+    opacity: 1,
+  },
+  out: {
+    opacity: 0,
+  },
+}
+
+const renderAnimatedLetters = string =>
+  Array.from(string).map((letter, index) => (
+    // eslint-disable-next-line react/no-array-index-key
```
jfranciscosousa, line 36: I disable this all together always. It's a rule that is rarely used and only usefull when dealing with animations and dynamic lists.

src/common/_fonts.scss
```diff
@@ -39,11 +39,12 @@
...
-    format("truetype");
+  src: url("../fonts/acta-display-extra-bold-no-tittle.woff2") format("woff2"),
```
jfranciscosousa, line 45: I did this to fix font jumping on the animation. If people don't have your fonts loaded beforehand the whole thing looks disgusting. Grappled with Baila about this and this was the solution that worked best.

jfranciscosousa: Hmmm, unicode-range though... I still think it would be best to inline this somehow. I can give you the tool I used

src/components/home/hero_section/title/animated_letters.js
pfac: This is the only instance of this warning in the codebase. I can disable it completely but I don't think it is worth right now to be honest

jfranciscosousa: Let it be then

config/plugins.js
```diff
@@ -1,10 +1,14 @@
...
+const postCssUrl = require("postcss-url")
 
 module.exports = root => [
   {
     resolve: "gatsby-plugin-sass",
     options: {
-      postCssPlugins: [postCssCustomMedia()],
+      postCssPlugins: [
+        postCssCustomMedia(),
+        postCssUrl([{ filter: "**/fonts/inline/*", url: "inline" }]),
```
jfranciscosousa, line 10: Does this inline stuff??????????????!!!!!!!!! :scream: It inlines the whole font still, I believe

src/common/_fonts.scss
pfac: I wasn't aware the inlining was to avoid a FOUT. Still the base64 representation of this font causes for an awfully long line.

Made some changes then:
* Added a PostCSS plugin to inline fonts based on path, allowing us to keep `base.scss` cleaner;
* Restored a smaller version of _Acta Display_ to be used in all titles with the most common English characters. The unicode range along with the `subset` word in the name of the font family should make that painfully clear;
* The regular font will be inlined to be loaded into the HTML. The tittleless version will be loaded normally since the splash should give us more than enough time to load it.

config/plugins.js
pfac: According to the plugin page that's what that does. :stuck_out_tongue: 

Hard to check in development, but testing it in the deploy preview https://deploy-preview-290--subvisual.netlify.com/, disabling cache there is no FOUT in the title. So seems like its working.

pfac: No javascript fall-back is screwed though, on it.

pfac: Fixed

src/common/_fonts.scss
jfranciscosousa: The inline can be a bit big. The base 64 only had the characters on the title.

jfranciscosousa: Anyway, seems ok

config/plugins.js
jfranciscosousa: 44k just the HTML doc, seems fine.

src/common/_fonts.scss
pfac: > The inline can be a bit big. The base 64 only had the characters on the title.

True, but:
- That inline did not replace _Acta Display_ completely, just in the homepage. This one replaces all ocurrences, including in blog posts where the content is dynamic, that's why I included more characters.
- The net result of removing _Acta Display_ and adding this one, without considering the inline version removed, is around -20KB.

config/plugins.js
pfac: Just for info purposes:
- New _Acta Display_ limited to the subset: 9.4KB WOFF + 7.4KB WOFF2 (this is what we inline);
- _Acta Display_ without tittle with only the `i`: 3.0KB WOFF + 2.2KB WOFF2 (this we load async).

We can inline only one of them to keep things even smaller. According to https://caniuse.com/#search=woff WOFF2 is only missing support in IE, but that would cost us only 2KB. What say you?

src/common/_fonts.scss
jfranciscosousa: Good

PR #295, opened 2020-04-07
Remove image placeholders on JS disabled

pfac:
Why:

* If Javascript is disabled all images loaded lazily are not shown
  because the placeholder never leaves.

### DIFF
--- a/src/components/image_loader/index.js
+++ b/src/components/image_loader/index.js
@@ -1,33 +1,37 @@
 import React from "react"
 import { useInView } from "react-intersection-observer"
 import Img from "gatsby-image"
+import _uniqueId from "lodash/uniqueId"
 
 import Placeholder from "../load_placeholder/placeholder"
-import useDetectJavascript from "../../utils/use_detect_javascript"
-
 import styles from "./index.module.css"
 
-const renderPlaceholder = (props, shouldRender) => {
-  if (!shouldRender) return null
-
-  return <Placeholder {...props} />
-}
-
 export default props => {
-  const hasJavascript = useDetectJavascript()
   const { delay, darkOverlay, imgStyle, ...imgProps } = props
   const [loaded, setLoaded] = React.useState(false)
   const [ref, inView] = useInView({
     triggerOnce: true,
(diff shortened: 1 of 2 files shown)


jfranciscosousa:
This ok, however, that hook is badly named. It should be `useIsMounted` and not `useDetectJavascript`.

But why? Because basically if we didn't have the splash screen, we would see the images **without** the placeholders. The javascript would need to be mounted for the browser to evaluate it and only after that would the image placeholders disappear.

The trick would be to render a `noscript` tag with inline styles, that just apply `display: none` to that element. I already do that somewhere, you could extract that behavior into a `HideInNoscript` component :joy:

pfac:
The hook name conversation is one we'll have to have later over coffee, already have some ideas in that regard. :stuck_out_tongue:

I actually considered that approach, and I'm torn between the two to be honest.

On one hand, the `display: none` approach pollutes the HTML with more than the browser actually needs if it does not have Javascript. And since, if it does, it rehydrates everything and React takes over, we could simply render the "no javascript" version in the server, which is both simpler to render and smaller to transfer, and let the rehydration bring the Javascript goodies to the party once it starts.

On the other hand, that approach gives the browser both versions and lets it do the choosing. And since 99% of the cases will have Javascript, this way the first rendering will already have all the elements in place. While nothing will actually work without React rehydrating everything, at least React will have less DOM changes to do.

I keep going back and forth between two approaches, unable to completely decide. So I'll take your vote as the tie breaker.

As for the wrapping element, I'm working on a feature to stop the animations on request, so I'll wait to analyse that idea since it might not fit afterwards.

jfranciscosousa:
Yeah, we can talk later. I prefer the latter version simply because it's the one that works better. It's perfect for noscript and perfect for with script. Even if the HTML is poluted. But we are using Gatsby anyway. The HTML will always be imcompreensible :joy:

jfranciscosousa:
You could however set a global rule for this kinds of stuff. You just have one `display: none` on the CSS and that's it. All instances of this component would then be hidden.

PR #299, opened 2020-04-09
Add share links in blog posts pages

pfac:
Why:

* As designed.

* Adding `react-share`, which seems to be the standard dependency to add
  social sharing links;
* Pre-computing the URL for each blog post after the Markdown node is
  created, relying on the `URL` environment variable which is set on
  Netlify to use the default site name or the custom domain the site was
  set up with (more info in the [Netlify docs]);
* Refactoring the wrapper components used in a blog post to accept a
  custom `className` from the parent components;
* Refactoring the blog post layout to place the sharing links on the
  left of the text, as designed, falling back to the vertical flow at
  the end of the post in mobile.

[Netlify docs]: https://docs.netlify.com/configure-builds/environment-variables/#git-metadata

### DIFF
--- a/src/components/blog/post/body/share_links.js
+++ b/src/components/blog/post/body/share_links.js
@@ -1,0 +1,47 @@
+import React from "react"
+import PropTypes from "prop-types"
+import classNames from "classnames"
+import {
+  FacebookShareButton,
+  LinkedinShareButton,
+  TwitterShareButton,
+} from "react-share"
+
+import useDetectJavascript from "../../../../utils/use_detect_javascript"
+
+import styles from "./share_links.module.scss"
+
+const BlogPostShareLinks = ({ className, url }) => {
+  const hasJavascript = useDetectJavascript()
+
+  // react-share links do not work with JavaScript disabled
+  if (!hasJavascript) return null
+
+  const rootClassName = classNames(styles.root, className)
+
+  return (
+    <div className={rootClassName}>
+      <div className={styles.content}>
+        <div className={styles.title}>Share</div>
+        <ul className={styles.links}>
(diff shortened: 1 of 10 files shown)


src/components/blog/post/body/share_links.js
```diff
@@ -0,0 +1,47 @@
+import React from "react"
+import PropTypes from "prop-types"
+import classNames from "classnames"
+import {
+  FacebookShareButton,
+  LinkedinShareButton,
+  TwitterShareButton,
+} from "react-share"
+
+import useDetectJavascript from "../../../../utils/use_detect_javascript"
```
jfranciscosousa, line 10: This pains me. One day try and refactor into using a `root` alias, not required for now

src/components/blog/post/body/share_links.js
```diff
@@ -0,0 +1,47 @@
...
+
+import styles from "./share_links.module.scss"
+
+const BlogPostShareLinks = ({ className, url }) => {
+  const hasJavascript = useDetectJavascript()
```
jfranciscosousa, line 15: Also, I was a bit stupid when I named this. A good alternative should be `hasMounted`, it's the step after the client rehydrates the server-rendered React page, so just keep that in mind

src/components/blog/post/body/share_links.js
pfac: What do you mean with `root` alias?

jfranciscosousa: You can use aliases on `yarn/webpack/gatsbyplugin` to reference the root of the project like `root/utils/use_detect_javascript`

PR #303, opened 2020-04-15
Fixing post listing and header PropTypes

naps62:
dates were being typed as `typeof(Date)`, which wasn't correct because they seem to always be passed around as a string, and only parsed on the Header component

retinaCover was marked as required, but very few posts actually have it

not sure if this the full fix, or even the correct one. happy to hear feedback

and why are we ignoring the non-retina cover altogether?

### DIFF
--- a/src/components/blog/post/header.js
+++ b/src/components/blog/post/header.js
@@ -46,8 +46,8 @@ BlogPostHeader.propTypes = {
   author: PropTypes.shape({
     name: PropTypes.string.isRequired,
   }).isRequired,
-  date: PropTypes.instanceOf(Date).isRequired,
-  retinaCover: PropTypes.string.isRequired,
+  date: PropTypes.string.isRequired,
+  retinaCover: PropTypes.string,
   title: PropTypes.string.isRequired,
 }
 
--- a/src/components/blog/post/header.js
+++ b/src/components/blog/post/header.js
@@ -46,7 +46,7 @@ BlogPostHeader.propTypes = {
   author: PropTypes.shape({
     name: PropTypes.string.isRequired,
   }).isRequired,
-  date: PropTypes.string.isRequired,
+  date: PropTypes.instanceOf(Date).isRequired,
   retinaCover: PropTypes.string,
   title: PropTypes.string.isRequired,
 }
--- a/src/components/blog/posts_list.js
+++ b/src/components/blog/posts_list.js
@@ -32,7 +32,6 @@ const renderItem = ({ frontmatter }) => {
     author: authorName,
     path,
     date: new Date(date),
-    slug,
     ...entry,
   }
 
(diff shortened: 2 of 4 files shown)


pfac:
I'm doing an experiment based on an article Zamith shared in our Slack today (https://chelseatroy.com/2019/12/18/reviewing-pull-requests/) where the author argues better than commenting is showing someone what you mean. Let me know if this bothers you, and please feel free to revert my commit if it does. :+1: 

> dates were being typed as `typeof(Date)`, which wasn't correct because they seem to always be passed around as a string, and only parsed on the Header component

This was an oversight on my behalf, but I was pursuing the opposite solution in #297. Instead of changing the dates back to strings, since the only reason they're strings is because GraphQL returns them as so, I thought they would be a lot more useful parsed into proper `Date` objects. See my suggestion in 2a08ac5.

> retinaCover was marked as required, but very few posts actually have it

Completely overlooked, nice catch. :+1:

> and why are we ignoring the non-retina cover altogether?

Because we can use Gatsby Sharp to generate all the versions we need from the retina version, so we won't be needing the non-retina version in the near future, therefore I didn't bother with it.

naps62:
@pfac thanks. I had read the article already, but while I do agree with a lot of what's said there, I have very strong feelings against the "push your own suggestions to the branch" part.
That makes sense for a lot of cases, but it directly interferes with the way we're used to work (aka: my ability to rebase and `push -f` my own branches after master updates). So, and without going too much into it here since it's offtopic, please don't do that :pray: creating a third branch and pointing me to the commit would work though, since I can cherry-pick it myself

As for this part:

> Because we can use Gatsby Sharp to generate all the versions we need from the retina version, so we won't be needing the non-retina version in the near future, therefore I didn't bother with it.

this way we're droping the cover images for all previous posts that didn't have the retina version. Was this a conscious decision? Why not just taht as a fallback when the retina one doesn't exist?

pfac:
> @pfac thanks. I had read the article already, but while I do agree with a lot of what's said there, I have very strong feelings against the "push your own suggestions to the branch" part.
> That makes sense for a lot of cases, but it directly interferes with the way we're used to work (aka: my ability to rebase and `push -f` my own branches after master updates). So, and without going too much into it here since it's offtopic, please don't do that pray creating a third branch and pointing me to the commit would work though, since I can cherry-pick it myself

Perfectly understandable. :+1: 

> this way we're droping the cover images for all previous posts that didn't have the retina version. Was this a conscious decision? Why not just taht as a fallback when the retina one doesn't exist?

I wasn't aware such posts existed, all the ones I saw had either no image or both. Please feel free to add the non-retina one as a fallback.

naps62:
@pfac my bad on the retina thing. apparently they all have them. I probably searched incorretly the first time

I'm fine with the changes you commit, but you need to be the one approving this, since I'm the author

PR #304, opened 2020-04-15
Moves posts to src/posts

naps62:
Post file names no longer reflect the URL they live under, so it's a bit
less useful to have them nested under `src/pages`.

Here, we move directly them to `src/posts`
```


naps62:
i'm used to merging my own stuff. But I got nothing against you merging the trivial stuff though, or even the not-so-trivial stuff if it become a blocker for something else

Especially on projects such as this, where we won't be available full-time to iterate

PR #305, opened 2020-04-15
Ability to reference filesystem images in markdown

naps62:
With this change, instead of creating a `my-post.md` file, we can create a structure:

```
my-post/
  index.md
  img.png
```

and inside, use regular markdown to include the image `badjerous`

This will allow us to publish images into our posts while keeping everything encapsulated in the same folder, and without us having to manually upload files to S3

PS: I tested this my creating a sample post with an image, using the structure mentioned above. Didn't include that test code here. But trust me, it works xD

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -53,6 +53,13 @@ module.exports = root => [
             width: 980,
           },
         },
+        {
+          resolve: "gatsby-remark-images",
+          options: {
+            maxWidth: 900,
+            linkImagesToOriginal: false,
+          },
+        },
         "gatsby-remark-external-links",
         {
           resolve: "gatsby-remark-prismjs",
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -56,7 +56,7 @@ module.exports = root => [
         {
           resolve: "gatsby-remark-images",
           options: {
-            maxWidth: 900,
+            maxWidth: 834,
             linkImagesToOriginal: false,
           },
         },
--- a/package.json
+++ b/package.json
@@ -19,9 +19,10 @@
     "gatsby-plugin-react-helmet": "^3.0.7",
     "gatsby-plugin-react-svg": "^2.1.1",
     "gatsby-plugin-sass": "^2.2.1",
-    "gatsby-plugin-sharp": "^2.0.22",
+    "gatsby-plugin-sharp": "^2.5.4",
     "gatsby-remark-embed-video": "^2.0.1",
     "gatsby-remark-external-links": "^0.0.4",
+    "gatsby-remark-images": "^3.2.2",
     "gatsby-remark-prismjs": "^3.4.1",
     "gatsby-remark-responsive-iframe": "^2.3.1",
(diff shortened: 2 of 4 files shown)


config/plugins.js
```diff
@@ -53,6 +53,13 @@ module.exports = root => [
             width: 980,
           },
         },
+        {
+          resolve: "gatsby-remark-images",
+          options: {
+            maxWidth: 900,
+            linkImagesToOriginal: false,
```
naps62, line 60: not sure about these two options.

I went with 900px since I saw that our posts have a max width of 850 or so.
the linking part just felt "right", but I'm open to feedback. By default, it creates an hover effect and a link to open the image on a separate tab

package.json
```diff
@@ -82,5 +83,8 @@
   "bugs": {
     "url": "https://github.com/gatsbyjs/gatsby/issues"
   },
-  "browserslist": [">0%", "not dead"]
+  "browserslist": [
+    ">0%",
+    "not dead"
+  ]
```
naps62, line 89: FYI: this was my own vim auto-intenting the file automatically

src/components/blog/post/body/_image.scss
```diff
@@ -1,15 +1,18 @@
 img {
+  margin: 40px 0;
+
+  @include media(">=desktop") {
+    margin: 56px 0;
+  }
+}
+
+img:not(:global .gatsby-resp-image-image) {
```
naps62, line 9: The attributes here were messing with the plugin's own styles, which introduces css classes, and some html structure to render a low-res preview and center the image.

So I changed this css to only use these positioning rules for images not rendered with the plugin. The margin part, I kept above because it seems to make sense for both

config/plugins.js
pfac: The max width in the design is set to 834px, we can bind to that value (I'll leave the suggestion).

As for the linking part, I agree with you. That is not an expected behavior in our blog from anyone in the team, and has never been mentioned as a need. I don't see any reason to include it. :+1:

config/plugins.js
```diff
@@ -53,6 +53,13 @@ module.exports = root => [
             width: 980,
           },
         },
+        {
+          resolve: "gatsby-remark-images",
+          options: {
+            maxWidth: 900,
```
pfac, line 59:
```suggestion
            maxWidth: 834,
```

src/components/blog/post/body/_image.scss
pfac: Thank you for the explanation. I was getting a bit triggered to be honest, mostly because I spent so much time aligning all of this. :sweat_smile:

Doesn't look good but I guess there's not much we can do since the plugin forces this on us. :/

package.json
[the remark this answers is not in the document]
naps62: I commented this exact line, saying it was my editor (probably some json autoformatting rule going on, don't know from where, and didn't seem too big a deal to go figure out)

you commented on the range though, which makes it appear as a separate thread

PR #306, opened 2020-04-16
New post - Continuous Stuff with Github Actions

naps62:
This is a transcription of a talk that I have at a few meetups. Had planned to made a written version for a while now

Pedro already gave me feedback on the equivalent PR for my personal blog (I'll be publishing this there as well, a few days later)

TODOs:
- Get more feedback
- Adjust image sizes


naps62:
@pfac looking at the preview for this one, I'd say I don't need to have that `HalfWidthImage` thing after all. Since the width here is smaller, it looks much better than my original one.

Can I haz your opinion?

naps62:
@pfac francisco wanted me to move the drop cap to the second paragraph (which is actually the first one of the actual post, since the initial text is just a prefix/notice)

do you have any better suggestion than what I did on this latest commit?

src/components/blog/post/body/_paragraph.scss
```diff
@@ -1,6 +1,7 @@
 @import "./variables";
 
-p {
+p,
+:global(.custom-paragraph) {
```
pfac, line 4: Instead of `custom-paragraph` how about `BlogPostParagraph` as a SuitCSS component? I get the idea, but in terms of styles, this is basically what you want, to be able to define that some element is to be seen as a regular paragraph. The fact we use the tags at all here are just a consequence of the Markdown rendering.

```suggestion
:global(.BlogPostParagraph) {
```

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,262 @@
+---
+id: 147
+path: /posts/continuous-stuff-with-github-actions
+title: "Continuous Stuff with Github Actions"
+author: miguel-palhas
+date: 2020-04-20
+tags:
+  - devops
+intro: "Last year, I took on the task of improving the continuous process over at Utrust. We weren't really happy with the amount of work that went into our releases, and I was looking for a more agile approach, where everyone from developers to the QA team could do their part with low friction."
```
pfac, line 9: You can turn this into a multiline string to prevent the wrapping in your editor and make it easier to edit. (https://yaml-multiline.info/)

```suggestion
intro: >
  Last year, I took on the task of improving the continuous process over at
  Utrust. We weren't really happy with the amount of work that went into our
  releases, and I was looking for a more agile approach, where everyone from
  developers to the QA team could do their part with low friction.
```

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,262 @@
...
+[circleci-orbs]: https://circleci.com/orbs/
+[orb-awscli]: https://circleci.com/orbs/registry/orb/circleci/aws-cli#orb-source
+[actions-checkout]: https://github.com/actions/checkout/tree/v2.0.0
+[twitter]: https://twitter.com/naps62
+
+<div class="custom-paragraph">
+<em>
+This post was adapted from a talk that I gave at several local meetups over the past few months. It got enough positive
+feedback that I felt compeled to convert it to written format. Original slides are available <a
+href="https://slides.com/naps62/continuous-stuff" target="_blanks">here</a>.
```
pfac, line 29: You don't need to use literal anchor tags to get the blank target, our posts will automatically render links like this (thanks to `gatsby-remark-external-links`, added in #286).

Also just noticed a typo: compel**l**ed :grimacing: 

```suggestion
feedback that I felt compelled to convert it to written format. Original slides are available [here][slides].
```

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,262 @@
...
+[primer]: https://www.imdb.com/title/tt0390384/
+[github-deployments]: https://developer.github.com/v3/repos/deployments/
+[github-deployment-status]: https://developer.github.com/v3/repos/deployments/
+[github-workflow-triggers]: https://help.github.com/en/actions/reference/events-that-trigger-workflows
+[circleci-orbs]: https://circleci.com/orbs/
+[orb-awscli]: https://circleci.com/orbs/registry/orb/circleci/aws-cli#orb-source
+[actions-checkout]: https://github.com/actions/checkout/tree/v2.0.0
+[twitter]: https://twitter.com/naps62
+
+<div class="custom-paragraph">
```
pfac, line 25: Forgot to add this suggestion to match the one in the stylesheets.

```suggestion
<div class="BlogPostParagraph">
```

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
naps62: the literal anchor tag is because I can't use markdown inside html tags

naps62:
FYI, converted to draft, because mendes is suposed to be publishing another post before mine

pfac:
@naps62 Mendes' post is live with ID 147. Yours is the next in line with ID 148.

naps62:
are those IDs being used at all, btw?

pfac:
> are those IDs being used at all, btw?

Yes, in the URL. There wouldn't be any clashing, but it would be weird. There's a plan to phase them out but involves redirect rules and I'm lazy.

naps62:
wait, the URL? I know we have that on the old posts, for backward-compatibility reasons, but doesn't the url come solely from the `path` given?
Or is `posts/148` also a valid url? and if so, why?

pfac:
> wait, the URL? I know we have that on the old posts, for backward-compatibility reasons, but doesn't the url come solely from the `path` given?
> Or is `posts/148` also a valid url? and if so, why?

- Because old posts had the ID in the URL, we're preserving the same URLs to keep the old links alive;
- We're keeping the old format until we move to something new;
- If we have so many posts with ID, it makes no sense (IMO) to change midway, better to keep things consistent and when we change we change them all together;
- Yes, the URL comes only from the `path` given.
- Which means you SHOULD choose an unique ID, and add it in the frontmatter fields, and use it in the path. But this is just so we keep things consistent until we change everything. Nothing breaks if you don't.

naps62:
| Because old posts had the ID in the URL, we're preserving the same URLs to keep the old links alive;

this is only valid for those old posts, and only means that we need to keep the number in the URL, not keep the ID field itself

| If we have so many posts with ID, it makes no sense (IMO) to change midway, better to keep things consistent and when we change we change them all together;

This seems to be the only "real" reason.
But I don't really see a point in keeping that, nor do I see a necessity of migrating (by implementing a redirect) the old posts to the new format. Having:
/posts/123-old-post
/posts/new-post-without-id

works perfectly. We can later think about adding `/post/old-post` as an option as well, but I don't see why that needs to be in place before discarding ids for new posts

pfac:
> this only means that we need to keep the number in the URL, not keep the ID field itself

True, however it would break consistency between past posts and new ones.

> This seems to be the only "real" reason.
> But I don't really see a point in keeping that, nor do I see a necessity of migrating (by implementing a redirect) the old posts to the new format. Having:
> /posts/123-old-post
> /posts/new-post-without-id
>
> works perfectly. We can later think about adding `/post/old-post` as an option as well, but I don't see why that needs to be in place before discarding ids for new posts

I'm aware it works, but it breaks consistency. It is a lot easier to mass clean these many files if all have the exact same format, than it is to deal with edge cases. But as I said, you can exclude the ID from your own post if you want.

Since this obviously bothers you, I'm bumping the priority of changing the posts paths. I'm already working on it. In the meantime, I'll leave it up to you whether you want to include the ID or not. I do prefer if you keep things consistent and since it makes the task easier.

pfac:
#318 submitted to address the ID issue.

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,268 @@
...
+id: 148
+path: /posts/continuous-stuff-with-github-actions
+title: "Continuous Stuff with Github Actions"
+author: miguel-palhas
+date: 2020-04-20
+tags:
+  - devops
+intro: >
+  Last year, I took on the task of improving the continuous process over at
+  Utrust. We weren't really happy with the amount of work that went into our
```
frm, line 11: Isn't it `utrust` now?

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,268 @@
...
+
+<div class="BlogPostParagraph">
+  <em>
+    This post was adapted from a talk that I gave at several local meetups over the
+    past few months. It got enough positive feedback that I felt compeled to convert it to written format. Original slides
+    are available <a href="https://slides.com/naps62/continuous-stuff" target="_blank">here</a>.
+  </em>
+</div>
+
+Last year, I took on the task of improving the continuous process over at [UTRUST][utrust]. We weren't really happy with
```
frm, line 37: Same as previous

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,268 @@
...
+the amount of work that went into our releases, and I was looking for a more agile approach, where everyone from
+developers to the QA team could do their part with low friction.
+
+Coincidentally, that project started right around the same time where GitHub Actions went into public beta. Some of our
```
frm, line 41: where or when? Does it make sense to go with `around the same time when GitHub Actions ...`

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,268 @@
...
+problems were related to shortcomings on our existing CI solution, so it felt right to see what GitHub had to offer.
+
+I ended up pleasantly surprised. But first, let's talk about what exactly was wrong with our previous CI.
+
+## Shortcomings of traditional CIs
+
+I've worked with a fair amount of CIs over the years...
+
+![CIs](./cis.png)
```
frm, line 50: idk if it's an issue with GitHub but this image isn't displaying for me.

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,268 @@
...
+                - staging
+```
+
+With this, commits to the `staging` branch would trigger a deploy to staging. Seems simple enough.  But now you need
+a branch for every single environment. And probably quite a few `push -f` commands, or similar git sorceries to force
+a commit from your normal workflow into a branch whose history is messier than the plot of [Primer][primer].
+
+## GitHub API to the rescue
+
+The problem here is that we're solving things the wrong way. A deploy is not a commit. It's a deploy.
```
frm, line 84: I always have this question: is it deploy or deployment? I thought the former was the verb and the latter the noun.

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
```diff
@@ -0,0 +1,268 @@
...
+
+## Manual workflows with comments
+
+We can also react to comments on the Pull Request itself, which can be useful if you want a more seamless integration of
+some features.
+
+In my case, I wanted to deploy a preview version of Pull Requests to our frontend application.  It wasn't efficient to
+do this for every single PR though (only a small subset of them actually need this), so we went with this instead:
+
+![Comments triggering GitHub Action](./github-actions-comment.png)
```
frm, line 158: Wait, what causes the trigger? Is it the 🚀 emoji? Is it a simple comment. Or the "preview" keyword?

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
naps62: it's Utrust now. It was UTRUST before

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
naps62: here it is wrong :+1:

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
naps62: actually it makes sense without either of them. `around the same time Github Actions went`

It's when, but it's implicit that way

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
naps62: really? to me it is showing
both on [github](https://github.com/subvisual/subvisual.com/blob/mp/ci-post/src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md) as well as in the preview on netlify

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
naps62: to be honest, I don't think there's a strict definition.

we also push "commits" and not "commitments"

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
naps62: uh. seems I accidentally deleted one sentence here. I distinctly remember writing it

Will add back. but FYI: it's the "preview" keyword. the emoji is the action answering back to acknowledge it got the message

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
frm: Ok, that's how I've been going on about it 👍

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
frm: I distinctly remember someone from their team correcting us and asking to use "utrust". The [website](https://utrust.com/) is inconsistent. Title says "Utrust", most sections use "utrust". 🤔

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
frm: Rich diff wasn't showing, it is now. Caching issue probs. 👍

src/posts/2020-04-16-continuous-stuff-with-github-actions/index.md
naps62: according to Jorge, Utrust is the name, utrust is the logo

PR #307, opened 2020-04-16
Fix lock file

pfac:
Why:

* #305 added `gatsby-remark-images` but used NPM instead of Yarn, adding
  the wrong lock file instead of changing the correct one. And the
  dumbass reviewer missed that.

* Installing the missing dependency using `bin/setup`, updating the
  correct lock file;
* Removing the wrong lock file.


jfranciscosousa:
I miss one PR and everything goes into flames :joy:

pfac:
> I miss one PR and everything goes into flames joy

That's how much I miss you.

PR #318, opened 2020-04-20
Remove IDs from blog posts

pfac:
Why:

* The first version of the Subvisual blog was created using Ruby on
  Rails. Back then, the path for each blog post was defined to be the
  [parameterize]d version of the blog post's title, prefixed by the
  post's numerical ID. The posts were stored in a database using
  auto-generated sequential numeric values.

  Each post has its own ID, but posts were not guaranteed to be
  published following this order, since the blog allowed for drafts.
  Drafts were just unpublished posts, but they had an ID to ensure
  persistence in the database.

  When the blog was migrated to GatsbyJS, the posts were all exported as
  Markdown files, retaining the ID of each post in the frontmatter.
  However this ID no longer has any purpose, neither as a field nor in
  the post's path.

* Removing the ID fields from all blog posts;
* Refactoring the existing components and blog post page generation
  logic to drop the use of the `id` field.

[parameterize]: https://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-parameterize

### DIFF
--- a/config/node/create_schema_customization.js
+++ b/config/node/create_schema_customization.js
@@ -15,7 +15,6 @@ module.exports = ({ actions }) => {
       }
 
       type Frontmatter {
-        id: Int!,
         author: Author! @link(by: "key"),
         date: Date!,
         path: String!,
--- a/src/components/blog/posts_list.js
+++ b/src/components/blog/posts_list.js
@@ -5,8 +5,8 @@ import Entry from "./posts_list/entry"
 
 import styles from "./posts_list.module.scss"
 
-const renderItem = ({ author, date, intro, id, path, title }) => (
-  <li key={id} className={styles.item}>
+const renderItem = ({ author, date, intro, path, title }, index) => (
+  <li key={index} className={styles.item}>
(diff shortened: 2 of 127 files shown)


netlify.toml
```diff
@@ -1,363 +1,363 @@
 [[redirects]]
-  from = "/posts/8-this-is-our-story/"
```
naps62: to be perfectly honest, I'd much rather keep the original urls with IDs for old posts, and keep the so-called inconsistency (I hesitate to call it that since it makes this sounds much worse than it is)

url's are ultimately not that important. the previous ones are readable as well, they just happen to have a now-meaningless number. It's not like anyone's reading the blog is gonna care

So, I can approve this, but I'd like to have someone else's input on this

pfac: Fair enough, requested a voting in Slack (https://subvisual.slack.com/archives/C02A7GR2N/p1587462683009100).

pfac: Voting closed, removing the redirects and keeping the `path` field.

PR #320, opened 2020-04-21
Adds optional SEO description to posts metadata

naps62:
meta descriptions should be at most 150..160 chars long (actual number varies according sources, and most just provide a range)
For post pages, we didn't have a description yet

This adds the option to do so in the post's frontmatter

a prop type validation in the SEO field is also introduced to prevent people from adding large descriptions

PS: the current validation uses 180 as the limit because that's the description we have on the blog index page itself. Ideally we would reduce that one as well, so we can change the limit to the 150..160 range

### DIFF
--- a/src/components/seo.js
+++ b/src/components/seo.js
@@ -75,6 +75,7 @@ SEO.propTypes = {
   lang: PropTypes.string,
   keywords: PropTypes.arrayOf(PropTypes.string),
   title: PropTypes.string,
+  description: PropTypes.string,
 }
 
 export default SEO
--- a/src/components/seo.js
+++ b/src/components/seo.js
@@ -75,7 +75,6 @@ SEO.propTypes = {
   lang: PropTypes.string,
   keywords: PropTypes.arrayOf(PropTypes.string),
   title: PropTypes.string,
-  description: PropTypes.string,
 }
 
 export default SEO
--- a/src/posts/2020-06-09-react-state-management-and-side-effects-with-redux-and-rxjs.md
+++ b/src/posts/2020-06-09-react-state-management-and-side-effects-with-redux-and-rxjs.md
@@ -3,6 +3,7 @@ path: /posts/react-state-management-and-side-effects-with-redux-and-rxjs/
 title: React state management and side-effects with Redux and RxJS
 author: gabriel-poca
 date: 2020-06-09
+description: Over the years I worked on many front-end projects. I don't always enjoy it, but I've found some libraries that make it more pleasant. One of those libraries is RxJS.
 intro: Over the years I worked on many front-end projects. I don't always enjoy it, but I've found some libraries that make it more pleasant. One of those libraries is RxJS.
 ---
 
--- a/src/posts/2020-06-09-react-state-management-and-side-effects-with-redux-and-rxjs.md
+++ b/src/posts/2020-06-09-react-state-management-and-side-effects-with-redux-and-rxjs.md
@@ -3,7 +3,6 @@ path: /posts/react-state-management-and-side-effects-with-redux-and-rxjs/
 title: React state management and side-effects with Redux and RxJS
 author: gabriel-poca
 date: 2020-06-09
-description: Over the years I worked on many front-end projects. I don't always enjoy it, but I've found some libraries that make it more pleasant. One of those libraries is RxJS.
 intro: Over the years I worked on many front-end projects. I don't always enjoy it, but I've found some libraries that make it more pleasant. One of those libraries is RxJS.
 ---
 
--- a/src/templates/blog/post.js
+++ b/src/templates/blog/post.js
@@ -75,19 +75,20 @@ const BlogPostTemplate = ({
 
 BlogPostTemplate.propTypes = {
   author: PropTypes.object.isRequired,
-  date: PropTypes.instanceOf(Date).isRequired,
-  html: PropTypes.string.isRequired,
   cover: PropTypes.string,
   coverFile: PropTypes.object,
+  date: PropTypes.instanceOf(Date).isRequired,
+  html: PropTypes.string.isRequired,
+  intro: PropTypes.string.isRequired,
+  seoDescription: PropTypes.string,
   title: PropTypes.string.isRequired,
-  description: PropTypes.string,
 }
 
 export default ({ data }) => {
   const { markdownRemark, coverFile } = data
   const { fields, frontmatter, html } = markdownRemark
   const { cover, url } = fields
-  const { author, date, title, seoDescription } = frontmatter
+  const { author, date, title, intro, seoDescription } = frontmatter
 
   return (
     <BlogPostTemplate
@@ -96,8 +97,9 @@ export default ({ data }) => {
         cover,
         coverFile,
         date: new Date(date),
-        seoDescription,
         html,
+        intro,
+        seoDescription,
         title,
         url,
       }}
--- a/src/templates/blog/post.js
+++ b/src/templates/blog/post.js
@@ -26,6 +26,8 @@ export const query = graphql`
           name
         }
         date
+        intro
+        seoDescription
         title
       }
       html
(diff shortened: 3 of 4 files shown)


src/components/seo/validations.js
```diff
@@ -0,0 +1,17 @@
+export const validateDescription = value => {
+  const maxLength = 180
+
+  if (!value) {
+    return null
+  }
+
+  if (!(typeof value == "string" || value instanceof String)) {
```
pfac, line 8: Do we really need the second condition? As far as I'm aware the first one is enough for strings. Also, shouldn't the first one be using strict comparison?

```suggestion
  if (typeof value !== "string") {
```

src/templates/blog/post.js
```diff
@@ -22,14 +22,22 @@ export const query = graphql`
...
 `
 
-const BlogPostTemplate = ({ author, date, html, retinaCover, title }) => (
+const BlogPostTemplate = ({
+  author,
+  date,
+  html,
+  retinaCover,
+  title,
+  description,
```
pfac, line 37: Do you mind keeping the alphabetic order? :grimacing: 

```suggestion
  author,
  date,
  description,
  html,
  retinaCover,
  title,
```

src/templates/blog/post.js
```diff
@@ -49,16 +57,30 @@ BlogPostTemplate.propTypes = {
...
 export default ({ data }) => {
   const { markdownRemark } = data
   const { frontmatter, html } = markdownRemark
-  const { author, date, retina_cover: retinaCover, title } = frontmatter
+  const {
+    author,
+    date,
+    retina_cover: retinaCover,
+    title,
+    description,
```
pfac, line 71: Do you mind keeping the alphabetic order? :grimacing: 

```suggestion
    author,
    date,
    description,
    retina_cover: retinaCover,
    title,
```

src/templates/blog/post.js
```diff
@@ -49,16 +57,30 @@ BlogPostTemplate.propTypes = {
...
   return (
     <BlogPostTemplate
-      {...{ author, date: new Date(date), html, retinaCover, title }}
+      {...{
+        author,
+        date: new Date(date),
+        html,
+        retinaCover,
+        title,
+        description,
```
pfac, line 82: Do you mind keeping the alphabetic order? :grimacing: 

```suggestion
        author,
        date: new Date(date),
        description,
        html,
        retinaCover,
        title,
```

src/components/seo.js
```diff
@@ -75,6 +77,7 @@ SEO.propTypes = {
   lang: PropTypes.string,
   keywords: PropTypes.arrayOf(PropTypes.string),
   title: PropTypes.string,
+  description: (props, propName) => validateDescription(props[propName]),
```
pfac, line 80: Why not:
- Use the validation function directly?
- Skip the `propName` argument since we already know the prop name here?
- Use `PropTypes.string` in this short function (or inside the validation function), and change the validation to check only the length of the string?

src/templates/blog/post.js
```diff
@@ -22,14 +22,22 @@ export const query = graphql`
         date
         retina_cover
         title
+        description
```
pfac, line 25: Do you mind keeping the alphabetic order? :grimacing:

src/components/seo/validations.js
naps62: string primitives and string objects are two different things in JS. Unless there's some `isString()` function that encapsulates this nowadays, this is the recommended way to do this check

naps62: the non-strict was an oversight. fixing

src/components/seo.js
naps62: 1. the function itself is both large enough and offtopic enough that I wanted to extract it to a diferent file
2. that makes the function unnecessarily brittle, and prone to mistakes if the name changes (which coincidentally, it will, given your suggestion of changing to `seo_description`). The argument is there already, so might as well use it, since there's no extra cost
3. not sure what you mean with either of these suggestions. can you provide samples? I don't know how to use `PropTypes.string` outside of assigning it to a prop directly

src/templates/blog/post.js
naps62: done

naps62:
> First, you're actively "solving" an issue but only for the most recent posts. I would very much rather clean up the intros that show up in the blog posts list, and use those as descriptions by default. I would use this description field as a way to override those intros if they're too long, for instance.

Anything along the lines of "let's fix this for all past posts instead of just improving new ones", I'll most likely be against.
I think I'm fine doing what you suggest, regarding using the intro as description instead of having a new field (and therefore trying to make sure it remains short enough). That has implications in the homepage content, so it's no longer just up to me, I'd say, but I would agree if the team agrees.
But retroactively changing the intros of all old posts to fit that, I'm out. That can be left to future work if someone feels it's important enough. But, it would a be a not-so-essential change blocking a much simpler and more importante one. 80/20 rule

> Second, these prop-types validations are not enforced. It requires the person writing the post to actively look into the console for errors. I understand we at least have some feedback like this, but it does not stop us from doing anything. I don't see the value.

My bad here. I thought this caused a compile time failure. Seems that the error only happens when using `gatsby develop` after all. I'll remove this part, and will see later if there's some other way to enforce this

> Also, I don't think description is obvious in the context of a blog post, maybe calling it seo_description would be better.

makes sense, will do :+1:

naps62:
ok on second thought I'm confused. do you want me to use the intro as description, or change the description to seo_description? those two suggestions seem incompatible

src/components/seo.js
pfac: Of course, my bad, should have included the suggestion in the end. Here's what I meant with point 3:

```javascript
  description: (props, propName) => PropTypes.string.isRequired(props, propName) || validateDescription(props[propName]),
```

You can also use `PropTypes.string.isRequired` inside the validation function, and change it here to:

```javascript
  description: validateDescription,
```

(This is what I meant with point 1, not inlining the function, sorry about the confusion :pray:)

Regarding point 2, what I meant is that you can simply assume the name of the prop that you already use in this very line. If it changes, you already have to change the line, I don't see how it is any more brittle. Example:

```javascript
  description: ({ description }) => validateDescription(description),
```

pfac:
> ok on second thought I'm confused. do you want me to use the intro as description, or change the description to seo_description? those two suggestions seem incompatible

Actually I don't have a fixed answer, because this was still on my _to think about_ list. I had ideas, but not a concrete approach I wanted to follow yet. That's why most of my comments were ideas for you to think about, not necessarily changes that needed to happen.

That said, since nobody wants to go through all the old posts, and add an SEO description, I think we should at least default to the first N characters of the intro as the description, and allow the authors to override the description through a new `seo_description` field.

In short, why not both?

```javascript
if (intro) return intro.slice(0, N)

return seo_description
```

naps62:
@pfac revisited this today. I dropped all the validation stuff, and went with an approach similar to yours.
Except that we don't need to manually truncate existing intros to 140 chars. That's what search engines will already do for us. This now just adds an optional `seoDescription` frontmatter field that will override the intro in terms of SEO

I guess the 140 limit will have to be enforced manually

src/posts/2020-06-09-react-state-management-and-side-effects-with-redux-and-rxjs.md
```diff
@@ -3,6 +3,7 @@ path: /posts/react-state-management-and-side-effects-with-redux-and-rxjs/
 title: React state management and side-effects with Redux and RxJS
 author: gabriel-poca
 date: 2020-06-09
+description: Over the years I worked on many front-end projects. I don't always enjoy it, but I've found some libraries that make it more pleasant. One of those libraries is RxJS.
```
pfac, line 6: Why this change?

src/templates/blog/post.js
```diff
@@ -76,13 +80,14 @@ BlogPostTemplate.propTypes = {
   cover: PropTypes.string,
   coverFile: PropTypes.object,
   title: PropTypes.string.isRequired,
+  description: PropTypes.string,
```
pfac, line 83: This should be `seoDescription` no? You're also missing `intro` here and below.

pfac: Also, can you please keep the alphabetical order? :pray:

src/posts/2020-06-09-react-state-management-and-side-effects-with-redux-and-rxjs.md
naps62: ah, forgot to remove this one apparently. my bad

PR #321, opened 2020-04-21
Removes cover. Renames retina_cover to cover

naps62:


pfac:
Do note that the build is failing.

PR #324, opened 2020-04-24
Migrating blog post 'Apprenticeship and Summer Camp Programs' by Laura

pfac:
Note: changed the quality of images placed with `gatsby-remark-images`
to 95%, since the default 50% was quite clearly screwing the images.


jfranciscosousa:
95 is way way too heavy. I find that most people won't notice the difference between 85 and 95, and 85 almost reduces size by half

pfac:
> 95 is way way too heavy. I find that most people won't notice the difference between 85 and 95, and 85 almost reduces size by half

Well, I could tell, but I'm aware of the change and I'll admit it is subtle. Let's go with 85%. Changed in 2178e86.

PR #327, opened 2020-04-26
implement shouldShowAnimation function

vladimirfomene:
I have implemented the shouldShowAnimation function which only shows animation on the homepage when you are just freshly visiting the site or you are visiting the site after 24 hours of your previous visit. @naps62 Please have a look and let me know how it can be improved.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -28,6 +28,30 @@ const spring = {
 
 const config = { spring, easings, getMargins: true }
 
+const shouldShowAnimation = () => {
+  const currentTime = new Date().getTime()
+
+  if (!window.localStorage.getItem("visited")) {
+    window.localStorage.setItem("visited", currentTime.toString())
+
+    return true
+  }
+
+  const visitedTime = Number(window.localStorage.getItem("visited"))
+  const ONE_DAY = 24 * 60 * 60 * 1000
+
+  if (
+    window.localStorage.getItem("visited") &&
+    currentTime - visitedTime < ONE_DAY
+  ) {
+    return false
+  }
+
+  window.localStorage.setItem("visited", currentTime.toString())
+
+  return true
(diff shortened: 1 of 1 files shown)


naps62:
FYI @pfac this is pretty much the code we did during the pairing session. It's a very rough draft though. It just hides the title entirely (including after the animation)

I need you or someone else to pitch in on what the correct approach should be, as I don't really know

pfac:
> FYI @pfac this is pretty much the code we did during the pairing session. It's a very rough draft though. It just hides the title entirely (including after the animation)
> 
> I need you or someone else to pitch in on what the correct approach should be, as I don't really know

@naps62 I actually don't know what the correct approach should be here :grimacing:. I'm currently working on changing how the splash screen animation works, so I'm not really sure if this should hold on until I'm done with it or if we should progress with it. I'm ok with progressing though, as I don't know how long it will take.

Either way, what I had envisioned here was more or less what you did, but maybe a layer below. As you said, this affects the hero title. That's because both components are connected to "sync" the animations. If you remove the splash, the title loses its trigger to appear. If you handled this below, the splash screen component would be the one deciding not to render, and could fire the title animation immediately.

Approach aside, some suggestions regarding the current code:
- We already have moment.js as a dependency of the project, so I suggest using it to write/read the timestamp from local storage, and for comparison of the two. I believe it would make the code simpler there.
- Consider using a [hook to access local storage](https://usehooks.com/useLocalStorage/). Should make the code a lot simpler, and more aligned with React standards.

Also, **one caveat**: the current splash screen was devised to be hidden if JavaScript is disabled, keeping the title visible with no animation. I don't see anything here that would break that behavior, just bear it in mind when changing things around.

naps62:
Closing in favor a potentially better solution

PR #328, opened 2020-04-30
Upgrade intersection observer polyfill

pfac:
`yarn outdated` says it's outdated. This fixes that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "gatsby-transformer-sharp": "^2.1.14",
     "gatsby-transformer-yaml": "^2.1.10",
     "include-media": "^1.4.9",
-    "intersection-observer": "0.7.0",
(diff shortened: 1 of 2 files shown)


jfranciscosousa:
Just a suggestion, dependabot works great with JS. A bit spammy though

pfac:
> Just a suggestion, dependabot works great with JS. A bit spammy though

Yeah that's the thing, when someone is available for this it is easy work, but dependabot will break my patience. Also, until we have some proper visual testing in place (working on it, but depends on the animations work), better not to automate this grunt work.

jfranciscosousa:
@pfac Usually updates don't change visual stuff. Usually, they break the build before that. But it's your choice. I only use dependabot on starter templates. Other projects just security updates and weekly linter updates. `yarn upgrade-interactive --latest` is great.

PR #329, opened 2020-04-30
Upgrade Prettier, ESLint, and their config

pfac:
Why:

* According to `yarn outdated` many of these are outdated, and since
  they have zero impact in the outcome of the website, there's no reason
  not to updated them.


.eslintrc
```diff
@@ -5,6 +5,7 @@
   ],
   "plugins": [
     "prettier",
+    "react",
```
jfranciscosousa, line 8: No need to use this as `airbnb` already imports it

pfac: I thought it was weird that plugin wasn't being used. Why the hell we need to have it as a separate dependency instead of havinb `eslint-config-airbnb` install it has its own dependency beats me. Fixed in 090ff8e.

PR #330, opened 2020-05-09
Add support for file paths in blog post covers

pfac:
Why:

* Following #305, where we added the ability to reference images in the
  local filesystem in the blog post Markdown body, our blog posts can
  now be self contained, with the required images sitting next to the
  blog post;
* That did not cover the blog post covers though (pun intended), which
  are expected to be URLs in the blog post frontmatter.

* Post-processing all nodes created by Markdown files and adding a
  `cover` node field, which is the cover URL or a path, resolving it
  into an absolute path from the Markdown file directory;
* Passing the new field to the page creation context, allowing to use
  this field to locate the image file in GraphQL;
* Refactoring the blog post template to extract the cover file sharp
  fragment and pass it down to the blog post header;
* Adding a blog post header cover component which renders a fluid cover
  if a file was found in the local filesystem, and keeps the current
  approach assuming an external URL otherwise.

Note: this change was devised in #324 and can be previewed in
https://deploy-preview-324--subvisual.netlify.app/blog/posts/apprenticeship-and-summer-camp-programs/

### DIFF
--- a/src/components/blog/post/header/cover.module.scss
+++ b/src/components/blog/post/header/cover.module.scss
@@ -1,0 +1,5 @@
+.root {
+  background-position: center bottom;
+  background-size: cover;
+  filter: grayscale(100%);
+}
--- a/src/templates/blog/post.js
+++ b/src/templates/blog/post.js
@@ -28,7 +28,7 @@ export const query = graphql`
     }
     coverFile: file(absolutePath: { eq: $cover }) {
       childImageSharp {
-        fluid(maxWidth: 980) {
+        fluid(grayscale: true, maxWidth: 980) {
           ...GatsbyImageSharpFluid
         }
       }
--- a/src/templates/blog/post.js
+++ b/src/templates/blog/post.js
@@ -11,29 +11,38 @@ import "../../common/base.scss"
 import styles from "./post.module.scss"
 
 export const query = graphql`
-  query($slug: String!) {
+  query($cover: String, $slug: String!) {
     markdownRemark(frontmatter: { path: { eq: $slug } }) {
-      html
+      fields {
+        cover
+      }
       frontmatter {
         author {
           key
           name
         }
         date
-        cover
         title
+      }
+      html
+    }
+    coverFile: file(absolutePath: { eq: $cover }) {
+      childImageSharp {
+        fluid(maxWidth: 980) {
+          ...GatsbyImageSharpFluid
+        }
       }
     }
   }
 `
 
-const BlogPostTemplate = ({ author, date, html, cover, title }) => (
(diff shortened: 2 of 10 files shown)


src/templates/blog/post.js
```diff
@@ -48,17 +57,26 @@ BlogPostTemplate.propTypes = {
...
-  const { author, date, cover, title } = frontmatter
+  const { markdownRemark, coverFile } = data
+  const { fields, frontmatter, html } = markdownRemark
+  const { cover } = fields
+  const { author, date, title } = frontmatter
 
   return (
     <BlogPostTemplate
-      {...{ author, date: new Date(date), html, cover, title }}
+      {...{
```
jfranciscosousa, line 72: This style of passing props is so weird 😅 fine by me tho

src/components/blog/post/header/cover.module.scss
```diff
@@ -0,0 +1,5 @@
+.root {
+  background-position: center bottom;
+  background-size: cover;
+  filter: grayscale(100%);
```
jfranciscosousa, line 4: You can grayscale images using sharp and save some processing power for the clients here

src/templates/blog/post.js
pfac: I prefer it because it is more succint and has the same readability, with no drawback, IMO ofc. The `prop={value}` notion in this case would be for the most part just repeating the same name for `prop` and `value`. Do you see any drawback in doing this?

src/components/blog/post/header/cover.module.scss
pfac: :thinking: nice catch, haven't thought of that. I'll look into it.

src/templates/blog/post.js
jfranciscosousa: Nope, it's just weird to me. I honestly would buckle everything into a `blogPostProps` object and then `...blogPostProps`. It's a bit more clear. But it's up to you

src/components/blog/post/header/cover.module.scss
jfranciscosousa: Not feasible on the Aurora website as they have filters and mix blend shenanigans, but if you are just grayscaling it, you can pre-process it. `filters` always take a bunch of power

jfranciscosousa: Don't forget to remove it here to test ;p

pfac: Added your suggestion of using sharp's `grayscale` in 6fd01f6. However I can't take this rule out yet as all cover images right now are external URLs, and are not being processed through sharp. If we get to move all images to the repository we can then remove this, but not until then.

PR #334, opened 2020-05-15
Tweak text width and blockquotes breakpoints

pfac:
Why:

* Reviewed with Baila the placement of the breakpoints in an effort to
  remove the interpolation and keep things a bit simpler;
* Reviewing the breakpoints, Baila asked for the text not to be as wide
  when the mobile font-size is used.

* Removing the SCSS `interpolate` function completely;
* Refactoring the blockquotes in blog posts body to align with the text
  when it does not have room for side paddings;
* Changing the width of the text when using the mobile font-size.


src/common/utils/_interpolate.scss
```diff
@@ -1,37 +0,0 @@
...
-
-/** Interpolate the values of a property in a range of the viewport width.
- *
- * The design only covers the minimum and maximum viewport sizes. While spaces
- * collapse to absorb the growth of the viewport, in some specific points the
- * rules are not set. So, instead of jumping straight from one position to the
- * other, this function allows us to gradually transition between the two
- * states.
- */
-@mixin interpolate(
```
jfranciscosousa, line 11: Yes die spawn from satan 😂

PR #340, opened 2020-05-29
Position Listing

zediogoviana:
List open positions from `YAML` file.

Why:

- Need to add and update open positions.

How:

- GraphQL to retrieve the list of open positions.
- Positions rendered in a responsive grid.


src/components/careers/academy_section.module.css
```diff
@@ -73,6 +80,21 @@
...
   }
 
+  .imgContainer {
+    position: absolute;
+
+    top: -10vh;
+    left: 0;
+
+    width: 100%;
+
```
gabrielpoca, line 90: what's the reason for these empty spaces?

zediogoviana: The linter enforces an empty line between some elements. But I checked now, and I have some that are not required! Will remove these ones right now.

PR #343, opened 2020-06-04
'How we set and track goals' by Laura Esteves

pfac:
Also adds a plugin to copy linked files to the site's public directory,
since GIFs are not processed by `gatsby-remark-images`.


jfranciscosousa:
image
is this supposed to look like this?

pfac:
I agree but the author approved it and prefers to keep that quote. :shrug:

PR #353, opened 2020-07-01
404 page (WIP)

naps62:
This is still a work in progress.

Currently looks like this:
image

What's missing:
- [ ] fix some sizes
- [ ] mobile / responsiveness
- [ ] different colors for nav links
- [ ] planets image


naps62:
I guess so

PR #354, opened 2020-07-08
Revert #352

pfac:
This reverts commit a8f817a.

Why:

* Blog post pages are now failing to open from the main blog page.


pfac:
Merged without approval to get the issue fixed in production as soon as possible.

PR #355, opened 2020-07-11
Refactor animations

pfac:


pfac:
The website is moving away from this code-base, this is no longer relevant.

PR #356, opened 2020-07-13
Adding Robz' post

naps62:


naps62:
I prefer to use a folder for all of them. if nothing else, to keep a more consistent ordering when looking at the directory (since files always show up after dirs)

PR #362, opened 2020-07-22
Transform files and folders in components to pascal case

gabrielpoca:
This change was mostly automated, so there are some places where we
should be using camelCase instead of PascalCase.
There are also a few places I didn't change because there was a naming conflict.
I intend to address those later, once I start touching the code itself. For now, I removed
that snake casing everywhere because this is a React project, component
files are named after the component, in pascal case.


pfac:
@gabrielpoca let's discuss this on Slack. There's a reason why the naming was like this, and there are several issues with doing this out of the blue.

gabrielpoca:
@pfac this was an automated change that I spent a couple of hours on. I intend to fine-tune some things after I can get this change merged, but. I don't want to do it all together both because it will be a mess, and also because I don't want to waste my time writing code that will stay in pull request forever. If we merge this, I know that there's an openness to make these changes.

gabrielpoca:
I can wait, but I think conflicts will be fine because of the nature of the change I'm doing.

pfac:
> @pfac this was an automated change that I spent a couple of hours on. I intend to fine-tune some things after I can get this change merged, but. I don't want to do it all together both because it will be a mess, and also because I don't want to waste my time writing code that will stay in pull request forever. If we merge this, I know that there's an openness to make these changes.

Sorry, I don't understand what you mean with fine-tune things. Both my questions regarding the code are for either a different rename, or for not renaming some of the files. Can you clarify where the fine-tuning comes in, so I can understand the impact of what I'm asking?

Also, can you share what you have in mind either on Slack or on Basecamp? That way we can coordinate better to avoid any frustrations.

> I can wait, but I think conflicts will be fine because of the nature of the change I'm doing.

Thanks :pray:. I understand that, it's just that many of these files are changed in those branches, and merging these changes in there will be harder than merging those changes into this one.

Since I know you don't want to go through the trouble, and I have a pretty good idea which components are safe to touch without affecting any other branch, would you oppose me splitting this PR to isolate the changes per component? Same end-result, less work for you. :innocent:

gabrielpoca:
@pfac by fine-tuning I mean move somethings around in order to organize in a typical react like structure. For instance, components are usually inside their folder with the `index.js` name, in this case, we have both a `Planets.js` and a `Plants` folder on the same level in the root. The other thing is that components should be named after the folder they are in, this also helps avoid nesting, which is strangely recommended (they are usually unopinionated) in React's docs. There are a couple of other things like this that would make the codebase be more components and less "templates". 

I'm ok with splitting, as long as I can merge some part, I know that I can continue with the changes. But did you try to rebase your branch on top of this one? It may not be that much trouble.

pfac:
> I'm ok with splitting, as long as I can merge some part, I know that I can continue with the changes. But did you try to rebase your branch on top of this one? It may not be that much trouble.

I did, there are a bunch of files I touched that Git becomes confused about, because after your change they no longer exist. I can't even start to fix that mess :confounded:

pfac:
@gabrielpoca why close this PR?

PR #363, opened 2020-07-23
Ability to specify a meta-image for posts without a cover

naps62:
Most of our posts nowadays don't have a cover image.  That leads to the preview
on social media being a default banner from the homepage, which tells nothing
about the post itself.

As a monkey patch, Laura and Baila have been manually creating images for each
post, after-the-fact (e.g.:
https://twitter.com/subvisual/status/1286280282354655234)

Not only is this laborious, it also works against social engagement, because
the tweet now consists of a link + image. The image itself is not a link to the
post, making the experience confusing (and frankly, I wouldn't be surprised if
we got less views over this)

This PR allows us to specify a `seoImage` in each post, which will be used for social media,
without the need for a cover existing on the post

(This will tie in nicely with the PR draft I'm about to open in a few seconds)

### DIFF
--- a/src/posts/2020-02-23-apprenticeship-and-summer-camp-programs/index.md
+++ b/src/posts/2020-02-23-apprenticeship-and-summer-camp-programs/index.md
@@ -4,6 +4,7 @@ title: Apprenticeship and Summer Camp programs
 author: laura-esteves
 date: 2020-02-23
 cover: ./cover.png
+seoImage: ./cover.png
 tags:
   - general
 intro: >
--- a/src/templates/blog/post.js
+++ b/src/templates/blog/post.js
@@ -54,35 +54,31 @@ const BlogPostTemplate = ({
   intro,
   seoDescription,
   seoImage,
-}) => {
-  console.log(cover)
-  console.log(seoImage)
-  return (
-    <Layout>
-      <SEO
-        description={seoDescription || intro}
-        image={seoImage || cover}
-        title={title}
-        url={url}
-      />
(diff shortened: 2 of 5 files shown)


src/templates/blog/post.js
```diff
@@ -52,35 +53,41 @@ const BlogPostTemplate = ({
...
-          </Wrapper>
-        </section>
-      </article>
-    </div>
-  </Layout>
-)
+  seoImage,
+}) => {
+  console.log(cover)
+  console.log(seoImage)
```
frm, line 59: 🚜

naps62: fixed

frm:
LGTM but get a second review from @pfac

src/posts/2020-02-23-apprenticeship-and-summer-camp-programs/index.md
```diff
@@ -4,6 +4,7 @@ title: Apprenticeship and Summer Camp programs
 author: laura-esteves
 date: 2020-02-23
 cover: ./cover.png
+seoImage: ./cover.png
```
pfac, line 7: Why specify `seoImage` when its value is the same as `cover` if we're using the cover automatically when this is not specified?

naps62: the build was failing, because there was no post at all with this field set at the moment, so gatsby couldn't infer the type.
I just fixed by using the new field once, in a backward-compatible way

PR #364, opened 2020-07-23
Generating social media banners with ImageMagick

naps62:
Why:

* Laura and Baila have been manually creating images for each post
  (e.g.: https://twitter.com/subvisual/status/1286280282354655234)
* That workflow probably doesn't produce the best tweets (see my previous post,
  just seconds ago), in terms of engagement

This PR adds a script that can be used to automatically generate
a banner similar to the one currently being generated manually.

It's not yet integrated into any CI workflow. It's a step towards that.

Also, the image generation itself is not yet finished. I need to tweak
font sizes and positioning


naps62:
Already done, and now allows me to provide an optional stock background image.

Sample banner generated with this script:

banner

gabrielpoca:
Looks fine. Why would `/tmp` bite you? I know a few people that use tmp for all kinds of temporary stuff, like downloading a repo just to test some things out. If I'm not mistaken, the /tmp is only cleared, in practice, when you reboot.

naps62:
* `mktemp` added. forgot about that
* banner.png removed. it was not suposed to be here at all
* `image_optim` call added at the end of the script

pfac:
> Looks fine. Why would `/tmp` bite you? I know a few people that use tmp for all kinds of temporary stuff, like downloading a repo just to test some things out. If I'm not mistaken, the /tmp is only cleared, in practice, when you reboot.

It's a global shared place, you might be interfering with other stuff running in the same machine without knowing. By using a `mktemp` the location should be only yours.

PR #366, opened 2020-07-24
New post: Unsucking your project onboarding

frm:


src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+```
+
+Sometimes, however, we need to do complex operations. As an example, in an
+internal project we had to set up `ngrok` before starting the server. Because
+this is something we tend to forget and newcommers might not be aware of, I made
+this `bin/` script that launches `ngrok` in the background, gets generated URL,
+launches the server and kills `ngrok` before exiting.
+
+```shell
+#!/usr/bin/env bash
```
naps62, line 106: you're missing the `-e` flag on all these scripts (fail on unhandled errors)

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+Webpack thingy. Wait, you used `npm`? Nope, not gonna happen, we use `yarn`.
+Yeah, run that again and delete this file. Is it done? Ok, you can run the app!
+
+It's not working? Oh yeah, the environment variables. We need them to start the
+app in development sometimes. Well, I'll send them to you through Slack even
+though Slack isn't encrypted.
+
+Ah, let's not forget setting up the database, migrating and seeding it. I guess
+that's all? Oh, how could we have forgotten... To use Phoenix you need to do the
+archive install. You're gonna need to do that as well.
```
naps62, line 37: "do the archive install" doesn't make sense, english-wise, unless you have the background to know the exact command to install the phoenix archive. might want to rephrase this

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+.
+
+.
+
+.
+
+Sound familiar? In **every** (and I mean every) project Subvisual does, we use
+`bin/` scripts. And not just to set up the project. **The programming languages
+and frameworks should be accessories to what you're doing. They're tools you use
+and making use of them should be as simple as possible.** In every project we
```
naps62, line 54: I honestly don't understand what this bold sentence is doing here. Seems completely out of context, when comparing to the previous & next ones

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+have a setup to do, a server to run and linters to check. So we abstract those
+steps with `bin/setup`, `bin/server` and `bin/lint`.
```
naps62, line 56: You forgot `bin/test`

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+
+This is particularly important for open source projects where you want people to
+join in and help with as little friction as possible.
+
+These first three scripts are so common, I have `s`, `se` and `l` aliased to
+`bin/{server,setup,lint}` just to avoid having to type all those extra letters.
```
naps62, line 62: not sure if I'm the only one doing it, but I do this: https://twitter.com/naps62/status/1279779629612568577

just a suggestion in case you want to incorporate it in the post. feel free to ignore

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+automate it.** This allows us to onboard people into the project by doing a
+simple `git clone` and `bin/setup`. **Time to first contribution is a couple of
+hours instead of 1+ days**.
+
+Let's get right into the basic scripts, the advanced scripts and the insane ones.
+
+## Into the basics
+
+As I mentioned, there are three essential scripts: `bin/server`, `bin/setup` and
+`bin/lint`. The first runs the development server. The second installs
```
naps62, line 74: again, `bin/test` missing here. might not be as common, but then again, it's usually a counterpart to `bin/lint`

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+everything you might need to run the project. The third runs all linters and
+checks, usually being equivalent to running the CI locally.
+
+I use these so much, both at Subvisual and on personal projects, I have aliases
+for all of them and I recommend you do the same.
```
naps62, line 79: deja vu?

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+
+### bin/server
+
+Usually the simplest script, these are examples for Elixir, Ruby and JavaScript
+(Gatsby) projects:
+
+```shell
+#!/usr/bin/env sh
+
+# Elixir
```
naps62, line 89: Phoenix would be more accurate

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+Anyone coming into the project doesn't need to be aware of this complexity, they
+would just run a command and voilá. Magic.
+
+### bin/lint
+
+Usually nothing more than the scripts you would run on CI. As an example, here's
+what we have on a project that runs the linters for frontend (in JavaScript) and
+backend (in Elixir). Notice the `set -e`. This causes the script to exit if a
+command has a non-zero exit code. The reasoning behind this is to mimic the CI
+behaviour.
```
naps62, line 163: so the `-e` is here after all? You know it's suposed to go on every other file as well right?

Also, `set -e` is equivalent to `#!/usr/bin/env sh -e`, ur just `#!/bin/sh -e` directly

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+
+Ah, the crown jewel. Usually the most complex script as it installs everything
+a developer might need. We usually make a few small assumptions:
+
+1. You have `postgres` installed. It's the database we use in almost every
+   project so 99.9% of the time we have it available in the system. Those 0.1%
+   of the time correspond to open source contributors or when we have a fresh
+   install (but we'll get into that). It's difficult to guess the way someone
+   might install `postgres` and for such a rare edge case, the script would get
+   unnecessarily complex.
```
naps62, line 199: to be fair, asdf-postgres is a thing, so you may want to suggest it here, instead of just dismissing with "postgres is difficult to install". In our case I agree with the 99.9% rule, but that's *our* case

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+the same.
+
+Notice that since `asdf` is not required, we check if it isn't installed and if
+not, we notify the developer and ask for confirmation. Otherwise, we dive right
+into the `asdf` parts.
+
+We run `set +e` before adding the plugins in case it fails since they're
+probably already added. Afterwards we'll need to import the `nodejs` release
+team keyring, before safely installing node, which is conveniently available on
+a `bin/` helper:
```
naps62, line 330: Now sure how long it's been since you read asdf-nodejs' readme, but this is not needed at all. They have this script themselves: https://github.com/asdf-vm/asdf-nodejs#install

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+variables we need. The condition makes sure no `.envrc` is present when doing
+the copy as to not override it. **We always try to make sure `bin/setup` is
+idempotent**, allowing others to run it repeatedly without making any damages.
+
+There have also been some variations of this snippet that check for the
+installation of `direnv`, similarly to what we do with `asdf`, but it has since
+been dropped.
+
+As a sidenote, if you are using `dotenv`, replace `.envrc` with `.env` and it
+should work!
```
naps62, line 399: FYI, there's also `git-crypt`, which is now my favorite option

again, a suggestion to mention. feel free to ignore

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+I previously mentioned the difficulty of managing secrets and environment
+variables. In projects that allow mock values for environment variables in
+development, we tend to put these fakes in `.envrc.sample`.
+
+Sometimes, however, we need to use sensitive data in development (e.g. API
+client ID and secrets). For those scenarios we make use of Keybase's encrypted
+git repos. They work just like a normal git repo, which means we can add them
+to our project as submodules. Usually we call the repo `secrets` and in them,
+files with the name of the environment they correspond to. So the `dev`
+environment would have a `dev` file and so on.
```
naps62, line 562: again, `git-crypt` :stuck_out_tongue:

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+several of us take this to a different level: abstracting the process of getting
+a new computer. **The philosophy is that your tool should be expendable. Having to
+set up a new one should be a matter of minutes or, at most, a couple of hours.
+Not days.**
+
+Personally, [I use an install script][dotfiles-install], 100% made in `zsh`. It
+installs and sets up [my dotfiles][my-dotfiles] and the apps I need.
+
+[Gabriel][gabriel-dotfiles] has a different philosophy, going with Ansible.
+[Naps][naps-terraform] uses Terraform and [Pedro][pedro-terraform] made a
```
naps62, line 756: I'd prefer `Miguel` here to be honest

naps62: and what the hell? I don't use terraform (where did that idea even come from?)
I'm old school and simple, I still use `rcup`

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+
+Occasionally, in the wild of our safari called Subvisual, we find other `bin/`
+scripts.
+
+### bin/test
+
+Found when the tests are complex or require multiple commands (e.g running
+integration tests separately, running frontend and backend tests). It's not that
+common to find these since most of us use setups with `vim-test`, similar to
+what [Naps uses for Elixir][vim-test].
```
frm, line 632: Miguel

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+tags:
+  - development
+  - devops
+intro: >
+  So you have a new developer joining your project. And, as usual, everything
+  changed in Javascript landscape since 37s ago. Now they have to install the
+  backend, frontend, all the tooling, etc etc etc. Stop this madness. The
+  programming languages and frameworks should be accessories to what you're doing.
+  They're tools you use and making use of them should be as simple as possible. At
+  Subvisual we have a system to handle this.
```
naps62, line 15: This intro is wayyyy too long
not just for visual reasons in our blog post list, but for SEO reasons as well. it will end up being cut down to the first ~140 chars, so you may want to rework this

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
+---
+path: /blog/unsucking-your-project-onboarding
+title: "Unsucking your project onboarding: bin/ scripts"
```
naps62, line 3: Is it just to me that the "bin/ script" part looks very confusing as part of the title?

pfac: I agree with Naps. I usually refer to these as utility scripts.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+source ~/.asdf/asdf.sh
+```
+
+As mentioned, we favour `asdf` for language version management. Usually this
+means we'll commit a `.tool-versions` in the repo to make everyone is running
+the same.
+
+Notice that since `asdf` is not required, we check if it isn't installed and if
+not, we notify the developer and ask for confirmation. Otherwise, we dive right
+into the `asdf` parts.
```
pfac, line 325: I don't think this is the case with most of our projects. We actually enforce these tools to keep everyone in sync when working on the project. Our scripts typically assume asdf-vm and fail when it is not present, as a clear way of saying _look, we use this, so either get on board or you're on your own_.

Not enforcing asdf-vm (or, back in the day, NVM or any Ruby version manager compatible with `.ruby-versions`) automatically removes the protection of everyone using the same versions for development.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+         pgp.mit.edu:80"
+
+for key in $KEYS; do
+  if ! gpg --list-keys $key 2>&1 >/dev/null; then
+    for server in $SERVERS; do
+      gpg --no-tty --keyserver "hkp://$server" --display-charset utf-8 --recv-keys "$key"  && break
+    done
+  fi
+done
+```
```
pfac, line 365: Why include the script at all? If you want to reference it, wouldn't it be better to just link to its source in the official repository? This is just unnecessary code the reader has to go through for no reason.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+
+```shell
+pp_info "setup" "Setting up the env..."
+
+if [ ! -f .envrc ]; then
+  cp .envrc.sample .envrc
+  printf "env file created, you might want to open .envrc and set the correct values. We recommend using direnv to manage it.\n\n"
+else
+  printf "envrc file already exists, skipping...\n\n"
+fi
```
pfac, line 384: As a reader I would be a lot more interested in seeing you load secrets from a secure source, like we have done using Keybase Git repos.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
pfac: TIL :+1:

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+sure `yarn` is available, installing Node dependencies, installing Elixir
+dependencies:
+
+```shell
+pp_info "setup" "Ensuring yarn is installed"
+if not_installed "yarn"; then
+  curl -o- -L https://yarnpkg.com/install.sh | bash
+else
+  echo "All done!"
+fi
```
pfac, line 413: You actually don't need to check for `yarn`, or install it from a remote script. Yarn is available as an NPM package. So you can instead just install Yarn globally (since we're using asdf-vm it is contained to the Node.js version we're using). If it is already installed, if will just be skipped, otherwise NPM will install it or update it for you.

```suggestion
pp_info "setup" "Ensuring yarn is installed"
npm install -g yarn
```

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+
+pp_info "setup" "Installing node requirements"
+yarn install
+
+pp_info "setup" "Installing elixir dependencies..."
+MIX_ENV=$env mix local.hex --force
+MIX_ENV=$env mix local.rebar --force
+MIX_ENV=$env mix deps.get
```
pfac, line 421: Why manually set `MIX_ENV`? At this point we already sourced `.envrc`, so wouldn't it make more sense to just have it there? Nevertheless, we didn't specify `NODE_ENV` for installing the JavaScript dependencies, and (as far as I'm aware) we don't use `bin/setup` for testing or production environments, so what's the point of setting the environment here if it defaults to development anyway?

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+mix archive.install hex phx_new 1.4.8 --force
```
pfac, line 422: Isn't this required only for generating new projects? If you're running `bin/setup` you're already inside an existing project, so isn't this a bit redundant?

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+pp_info "setup" "Installing python requirements"
+pip install -r requirements.txt
+
+# Gatsby project (after ensuring the yarn install)
+pp_info "setup" "Ensuring gatsby is installed..."
+if not_installed "gatsby"; then
+  npm install -g gatsby-cli
+else
+  echo "Already installed, skipping."
+fi
```
pfac, line 442: Same reasoning as for Yarn above (you can simply install it and it will be skipped if already installed). However this is only needed for generating new projects, so the same reasoning as install `phx_new` applies here.

```suggestion
npm install -g gatsby-cli
```

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+a new computer. **The philosophy is that your tool should be expendable. Having to
+set up a new one should be a matter of minutes or, at most, a couple of hours.
+Not days.**
+
+Personally, [I use an install script][dotfiles-install], 100% made in `zsh`. It
+installs and sets up [my dotfiles][my-dotfiles] and the apps I need.
+
+[Gabriel][gabriel-dotfiles] has a different philosophy, going with Ansible.
+[Naps][naps-terraform] uses Terraform and [Pedro][pedro-terraform] made a
+monster that has CI checks and is completely provisioned through Docker.
```
pfac, line 757:
```suggestion
monster that has CI checks and is completely provisioned through Ansible.
```

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+else
+  set +e
+  asdf plugin-add erlang https://github.com/asdf-vm/asdf-erlang.git 2>/dev/null
+  asdf plugin-add elixir https://github.com/asdf-vm/asdf-elixir.git 2>/dev/null
+  asdf plugin-add nodejs https://github.com/asdf-vm/asdf-nodejs.git 2>/dev/null
+  set -e
+
+  printf "Importing nodejs release team keyring... "
+  bin/helpers/import-nodejs-keys
+  echo "Done!"
```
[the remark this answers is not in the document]
frm: I disagree with this, in the sense that I'd rather have a developer not having to deal with this and quickly join in than having to go through those battles themselves. I really disagree with the philosophy of "let's make it difficult just because, since that way they'll learn".

Either way, I'll reference the script provided by `asdf-node`

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+
+  asdf install
+fi
+
+pp_info "setup" "resourcing environment"
+source ~/.asdf/asdf.sh
```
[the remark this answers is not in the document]
frm: If you install the languages for the first time, they won't be available and the following commands will fail.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
frm: Yes, but it's necessary if you have umbrellas. My take was just "leave it be, can't harm"

src/posts/2020-07-27-unsucking-your-project-onboarding.md
frm: IIRC this came from a certain collab with the office in front. `MIX_ENV` had to be changed to debug certain issues. I found it to be needed as well in personal projects.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
frm: Forgot to delete when I moved it upwards

src/posts/2020-07-27-unsucking-your-project-onboarding.md
naps62: tried, but not easy to do on build time.
currently enforcement is done in code reviews only

this is not a hard limit, since 140 chars is not a strict limit. Just a general commendation, since it's usually the maximum size you'll see in a lot of smaller devices or contexts

src/posts/2020-07-27-unsucking-your-project-onboarding.md
naps62: Actually, it *can* harm

You're hardcoding a Phoenix version deep in the setup scripts. pretty easy to miss if phoenix is upgraded in the project

frm:
> I personally feel you're all over the place with this one.

I know this is your feedback, but I respectfully disagree. I actually took my time to review how I wanted a more "declarative" post (as opposed to the storytelling style I usually go for). The format is meant to serve as a page to consult repeatedly and I'm quite happy with the outcome.

> When you describe bin/setup, why comment each section of the script individually, only to repeat the whole script later?

Because I've genuinely grown to hate when people breakdown a script and don't make it available in full afterwards. I also dislike when a giant script is posted and then described without visual references. I hate breaking my reading flow having to scroll up and down and I don't want to impose that on people. Unlike me, if they're comfortable with scrolling up and down, they're comfortable with scrolling past the full script.

> Why link your ASCII cinemas to another page instead of embedding? Less jumps make happier readers.

They're embedded through JS, they're not videos. I tried to embed but it's not that simple. Are we scrubbing script tags?

> (not a question) I would drop all the alternative language examples. The message is language agnostic, the examples are not, but any intermediate user should be able to know what commands to use for their language/framework if you tell them what we aim to accomplish there. That way you can simply use Elixir or Javascript examples, and leave Python/Ruby/whatever out of the speech.

Yeah, I'm comfortable with dropping those.

> (also not a question) You already do a good job of explaining why these scripts are useful, I don't think the list of examples at the end has any value. I also don't think bin/lint is a good example to feature. bin/console or bin/test or bin/deploy all have would have more value for me as the reader.

I don't have access to a valuable `bin/console` anymore, unfortunately. Why wouldn't `bin/lint` be a good example?

> (again, not a question) I recommend completely removing your bin/functions of these examples. Maybe quickly refer in the end we sometimes extract common functionality. But for anyone learning about these scripts, make their lives easier and just stick to simple and short things. If you want to showcase bigger things like your Phoenix over Ngrok script, put it in a Gist and link to it, don't force readers who are not as advanced (or who don't care about Ngrok) to have to go through it.

I disagree as they have cute little touches that readers might find in the asciinema. As for "force readers who are not as advanced", I quote "Readers can scroll both ways 😉"

> Why mention our dotfiles and install scripts? Not that I didn't enjoy the compliment, I'm always eager to boast about my oneliner-setup-that-never-works-when-I-actually-need-but-its-still-awesome. I just feel you dilluted the message there of preparing a project to make everyone's lives easier by automating simple boring tasks.

Because the philosophy that applies to the projects applies to computers. This is the point of the blog post. Languages are tools, frameworks are tools, they change over time, we should make onboarding as simple as possible. We take this a whole new level. Computers are also tools and also expendable, we should make onboarding as simple as possible.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
frm: Actually true. Any recommendation other than deleting? Maybe getting the correct phoenix version from `mix.exs`?

src/posts/2020-07-27-unsucking-your-project-onboarding.md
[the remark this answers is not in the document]
pfac: Rails projects also didn't use to have a `bin/lint`, we would simply have `rake lint` (or `rake rubocop`) and `yarn run lint`. Not that I mind a `bin/lint`, just that the task runner was usually able to take care of it.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
```diff
@@ -0,0 +1,789 @@
...
+I use these so much, both at Subvisual and on personal projects, I have aliases
+for all of them and I recommend you do the same.
+
+### bin/server
+
+Usually the simplest script, these are examples for Elixir, Ruby and JavaScript
+(Gatsby) projects:
+
+```shell
+#!/usr/bin/env sh
```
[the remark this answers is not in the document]
pfac: > If you run `chsh` does it change to `zsh`? I believe so because this wasn't working with `bash` in a script I ran, but with `zsh` it was.

I believe it doesn't change, but that's something system dependent. On Linux it doesn't (otherwise I wouldn't have to care about `dash` on Ubuntu since I always used Bash/Zsh).

> Either way, I've been updating my scripts to be `zsh` or `bash` dependent, so I'll run with that.

That's a valid approach :+1:

src/posts/2020-07-27-unsucking-your-project-onboarding.md
pfac: I'm sorry I don't get the argument. If you change the project's Node version, don't you re-run `bin/setup`? Also, if you look at the install script the "agnostic setup" does, that's the tarball of the project decompressed into `~/.yarn`. You're polluting the user's home directory instead of installing a tool in the contained environment set by asdf-vm.

I have nothing against someone preferring the "agnostic setup", but I would be utterly pissed off if I started working on a project and suddenly it started installing random stuff in my home folder without asking.

src/posts/2020-07-27-unsucking-your-project-onboarding.md
pfac: I still don't get it...

src/posts/2020-07-27-unsucking-your-project-onboarding.md
pfac: > Yes, but it's necessary if you have umbrellas. My take was just "leave it be, can't harm"

I don't get it, doesn't this _if_ actually make this a very project specific thing? Wouldn't it make more sense to keep it out of a generic example and just use it in umbrella apps?

src/posts/2020-07-27-unsucking-your-project-onboarding.md
pfac: I don't understand you're point, does this mean you agree it is unnecessary, or do you believe it is necessary?

src/posts/2020-07-27-unsucking-your-project-onboarding.md
frm: And installing globally is fine then?

src/posts/2020-07-27-unsucking-your-project-onboarding.md
naps62: deleting would be my option.

especially because, now that I think about it, this isn't needed for onboarding at all. it's only needed in the very specific case where:
1. you're a new guy, that didn't have phoenix installed
2. your project is an umbrella
3. you want to add a new phoenix app as a child to this umbrella
4. you went through enough time at this project that you're comfortable adding a new phoenix child, yet not as much time as to have installed phoenix yourself. do these two even overlap?

pfac:
> I know this is your feedback, but I respectfully disagree. I actually took my time to review how I wanted a more "declarative" post (as opposed to the storytelling style I usually go for). The format is meant to serve as a page to consult repeatedly and I'm quite happy with the outcome.

It is just my opinion, please feel free to dismiss it :smile:. A lot of my feedback here is precisely because I understood you wanted this to be a reference article for future apprentices/summer campers/learners to refer to and understand the value of automating these tasks, and how to do it. That's why I went to the trouble of commenting so many things on how to make it simpler and more generic.

Of course, you're free to disagree and do whatever you want, it's your article after all. I would just ask you to think what's your opinion and what's something we do collectively as a team, and make it clear when its one or the other. Otherwise you might be misleading the reader, which might not be working directly with you. :peace_symbol: 

> Because I've genuinely grown to hate when people breakdown a script and don't make it available in full afterwards. I also dislike when a giant script is posted and then described without visual references. I hate breaking my reading flow having to scroll up and down and I don't want to impose that on people. Unlike me, if they're comfortable with scrolling up and down, they're comfortable with scrolling past the full script.

Ok, we simply have different opinions on this. Nothing wrong with that. I suggest then just moving the entire script to a gist, to remove the repetition. One thing I appreciate in posts like this is having the scripts quickly accessible at the top, so I can revisit them quickly through the post. Maybe gather all the examples you use here in a single gist and link the reader at the top?

> They're embedded through JS, they're not videos. I tried to embed but it's not that simple. Are we scrubbing script tags?

We're not. You can see how I embedded it when migrating your Jobs and Timers post from Medium here: https://github.com/subvisual/subvisual.com/blame/master/src/posts/2019-06-07-jobs-and-timers-in-neovim-how-to-watch-your-builds-fail.md#L58

> I don't have access to a valuable `bin/console` anymore, unfortunately. Why wouldn't `bin/lint` be a good example?

Just because it is very common to simply do it with each language's task runner. The most value I've seen in it is to automatically go through all the linters in the project, which is very commonly just a couple commands. In fact when the linters have auto-fix capabilities you often use the command directly to enable such things.

`bin/console` is added by default on any new gem created with Bundler, for instance. I've seen more value in it because it allows you to not only configure things like the prompt, but it also allows you to prepare some things automatically. For instance, in development you can use to automatically prepare some data to make it easier for you to test things out.

`bin/deploy` in my opinion would probably be an even better example, if you don't mind touching a bit of its own quirks. It can be as simple as ensuring the Git branches are set up and pushing to Heroku. It has the advantage that this script can actually hide from a regular developer details about the running machines they might not care or not be aware. One day the production machine is Heroku, the next is Digital Ocean, most developers won't care if all they have to do is `bin/deploy`.

Again, this is just food for thought. I just found odd the use of `bin/lint` because I rarely see it. You do you. :heart: 

> I disagree as they have cute little touches that readers might find in the asciinema. As for "force readers who are not as advanced", I quote "Readers can scroll both ways wink"

I honestly didn't understand what you mean with the quote. My point was just that you use things like `pp` which are not standard shell commands, and while they have some value in a project, they don't bring anything to the message you want to convey here. I'm just suggesting keeping things simple(r).

> Because the philosophy that applies to the projects applies to computers. This is the point of the blog post. Languages are tools, frameworks are tools, they change over time, we should make onboarding as simple as possible. We take this a whole new level. Computers are also tools and also expendable, we should make onboarding as simple as possible.

This one caught me off guard. I thought the point of the blog post was to teach and showcase the value of utility scripts in a project, not automation in general. I don't think the message is that broad. But again, that's my opinion. :peace_symbol: :v:

src/posts/2020-07-27-unsucking-your-project-onboarding.md
pfac: "globally" here applies to the Node.js version you're running only. If you change it, it goes away. If you delete it, it goes away without influencing any other version. Any other tool I would even argue we don't need to install it globally, because we can simply install it as a project dependency and run it as `npx tool`. But since Yarn installs the dependencies, it has to be "global" just like NPM (which with asdf-vm is also Node.js version dependent).

I might be reading too much into your message but I'm starting to get a vibe that you're feeling attacked and I really had no intention of it. Merely wanted to be as helpful as I could. Please proceed as you see fit. :pray:

frm: But npm is bound to node versions, yarn isn't. It's supposed to act as an independent tool, hence they recommending that option.

pfac: But that's not the option they recommend, it's precisely to install Yarn globally with NPM. https://yarnpkg.com/getting-started/install#global-install

PR #372, opened 2020-08-24
Refactor SEO component and fix its image

pfac:
Why:

* Our blog posts are currently not being rendered with the correct SEO
  image URL because we seem to be resolving the file path locally but
  never converting it to the final URL.
* The SEO component is also rather hard to grasp, with lots of tiny bits
  of logic to handle the provided information as props with proper
  fall-back on the defined site metadata as defaults.

This change addresses the issues by:

* Refactoring the SEO component, extracting the query to a hook, and the
  many groups of meta-tags to their own sub-components handling their
  own logic; the root component is left in charge of merging the
  provided metadata with the site's defaults;
* Refactoring the blog post template to use gatsby-image fragments to
  resolve the SEO image by its absolute path, thus fixing the final URL
  in the rendered page.

Before:

Screenshot from 2020-08-28 17-25-34

After:
Screenshot from 2020-08-28 21-47-17

NOTE: Since we're fixing some of the conventions for this project, I decided to use the `.jsx` extension for JSX files, but turns out `eslint-plugin-import` does not deal well with importing component `X` when its source file is `X/index.jsx`. Spent quite a while researching how to go around this or reconfigure the plugin with no success. Ended up deciding to disable it. Additionally, I noticed the `import/no-unresolved` rule had some misconfiguration and was also being applied to local imports, which is completely redundant since the GatsbyJS build process will already verify that. One could argue the linter provides a good tip while editing the file, but the truth is that `eslint-plugin-import` is prepared to use Webpack to resolve imports, but only when it has access to the configuration file, which we don't with GatsbyJS.


src/components/SEO/MetaTags.jsx
```diff
@@ -0,0 +1,38 @@
...
+  description,
+  image,
+  lang,
+  keywords,
+  title,
+  twitter,
+  url,
+}) => (
+  <>
+    <Helmet htmlAttributes={{ lang }} title={title}>
```
jfranciscosousa, line 19: Why not just put everything into a single component, splitting this off is just creating entropy.

src/components/SEO/MetaTags.jsx
```diff
@@ -0,0 +1,38 @@
...
+      <meta name="description" content={description} />
+      <meta name="title" content={title} />
+    </Helmet>
+    <Keywords keywords={keywords} />
```
jfranciscosousa, line 23: Why not just put everything into a single component, splitting this off is just creating entropy.

src/templates/blog/post.js
```diff
@@ -40,9 +42,23 @@ export const query = graphql`
...
+        fixed(width: 2160, height: 1080) {
+          ...GatsbyImageSharpFixed_noBase64
+        }
+      }
+    }
   }
 `
 
+const resolveImageURL = ({ file, src }) =>
+  pathToURL(_get(file, "childImageSharp.fixed.src") || src)
```
jfranciscosousa, line 56: _get 3rd argument is the default value, you can use that.

src/components/SEO/MetaTags.jsx
pfac: Why do you say that?

Everything is currently in the same single SEO component, which is making things very hard to read and change IMO, because there's all this logic and tags floating around. In this specific case, by handing off the keywords we can make it optional (not render the tag) without having to pollute this level with branching in JSX.

IMO, this is actually reducing entropy by making things easier to read, understand, and more orderly (one component with the single responsibility of handling keywords). Nonetheless this is my opinion, would like to understand yours.

src/templates/blog/post.js
pfac: True story, my bad :+1:

src/components/SEO/MetaTags.jsx
jfranciscosousa: It's ok. I just don't split components anymore this way, at all. Being easier to read is debatable. You have to open multiple files to understand what is going on. For logic in JSX you can just `condition && <Something/>` that for me is easier to read. I just do this because on the frontend you are always writing a bunch of stuff and if you keep with this pattern you might end up with thousands of components, where most of them are just used in one place.

I only split things off when they need to be reused on more than one place, or to isolate them for testing purposes. Or just have a single file with more than one component or render functions.

pfac: I agree that's debatable. I don't consider the `condition && <Something />` easy to read, especially when it starts spawning multiple lines. I understand what you're saying about splitting components when they're to be used in more than one place, but in these cases you have a top-level component that's being reused, and you have nested components inside this one's directory that are meant to be used by it. Yes, you might get a lot more components, but the structure itself tells you how they're organized (the hierarchy).

As for the argument of a single one with multiple components/render functions, while I have nothing against it per se, I do refrain from doing it due to the length the files end up having. I would rather have hundreds of little files, with very easy to understand and self-contained components, then to have a single file spawning over 100 lines where you get lost easily. It's like taking the single-responsibility rule but to the file level, each file is in charge of one component so sub-components go to nested files.

Difference of opinion I reckon. :man_shrugging:

jfranciscosousa (approved):

PR #373, opened 2020-08-24
Setup Ruby with ImageOptim for SEO image generation

pfac:
Why:

* The SEO image generation script uses the `image_optim` Ruby gem, but
  this project does not set it up by default.


naps62 (approved):

frm:
Out of curiosity, why was the Ruby downgrade needed?

pfac:
Apparently Netlify reacts to the fact it has Gemfile in the repository by attempting to install these dependencies, and it is using Ruby 2.6.2 by default.

frm:
Wait, so we downgrade the Ruby version instead of updating the version used by Netlify? It's a single file: https://docs.netlify.com/configure-builds/manage-dependencies/#ruby

pfac:
> Wait, so we downgrade the Ruby version instead of updating the version used by Netlify? It's a single file: https://docs.netlify.com/configure-builds/manage-dependencies/#ruby

What's the problem? There's nothing Ruby related even being used in Netlify, this is just so we can version `image_optim` so the Ruby version just has to be a supported one, and we don't use `.ruby-version` with asdf-vm so this would have to be done via environment variable (which we don't need for the previous reasons). There's absolutely no impact caused by this downgrade aside using a different number.

PR #374, opened 2020-08-25
Extract generic utilities to @subvisual/utils

pfac:
Why:

* Some of these utilities are used in configuration files where we can't
  use ES6 features, and have to rely on Node.js-style CommonJS modules,
  forcing us to keep a second style for these utilities.

  Additionally, it's also not clear how to organize these utilities,
  since we added an import alias to avoid having to keep relative paths,
  but in the configuration files this alias is not yet usable.

  Extracting these generic utilities to an external package allows us to
  reuse them in other projects, and is a bit less code we have to
  maintain in this repo specifically.


naps62:
Looking at the package itself, and its usage here, I have to just question *why*?

It seems the only problem mentioned was the fact that they had to be CommonJS, which doesn't really sound like a real issue to me.
And looking at the codebase after this change, I will now be wondering what the hell does `normalizePathForRegExp` do, since the name won't be very helpful, unlike lodash's functions that are mostly self-explanatory

And this:

> Extracting these generic utilities to an external package allows us to
reuse them in other projects

Is for me a symptom of premature optimization

pfac:
> Looking at the package itself, and its usage here, I have to just question _why_?

This has many reasons actually, the most important for me is that we've all added generic tools to most our projects, and we keep not preserving useful stuff. My motivation for creating `@subvisual/utils` is the same I had for creating the `subvisual-utils` gem, to try to improve our lives a bit.

> It seems the only problem mentioned was the fact that they had to be CommonJS, which doesn't really sound like a real issue to me.

For me the pain of having to juggle where these things are stored is a massive one. By putting this in a package I can stop worrying where to put it, I just need a generic function for X, best push it to the utilities package.

> And looking at the codebase after this change, I will now be wondering what the hell does `normalizePathForRegExp` do, since the name won't be very helpful, unlike lodash's functions that are mostly self-explanatory

:confused: The name is pretty self-explanatory to me, at most I can understand one wouldn't be sure what _normalization_ means in this case, you would have that pain before. I'm more than open to better suggestions for the name though :smile:.

Otherwise the code is still accessible, even if in a different repository. Would it help to have it as a submodule? Not a fan of the approach, but if it's better for you :shrug:.

> 
> And this:
> 
> > Extracting these generic utilities to an external package allows us to
> > reuse them in other projects
> 
> Is for me a symptom of premature optimization

I got to ask why? The pains this solves are not new (at least for me), and I've had them in other projects. Whether someone has use for these particular functions in another project I honestly don't care, but I would have appreciated to have them available before. Call it the "premature optimization" I wish I had at the time.

naps62:
My comment is 100% based on the PR description. If you and other people have felt the need for this before, it's a slightly different matter, but that's not what I got from the original description

I don't agree with the idea that just because we've used a function a few times in different projects, that a generic "subvisual utils" library is a good idea. Looking at the two functions currently on this one, I stand by my opinion. Have never looked at the Ruby one, to be honest.

> For me the pain of having to juggle where these things are stored is a massive one. By putting this in a package I can stop worrying where to put it, I just need a generic function for X, best push it to the utilities package.

I myself would simply avoid using such a package, even if maintained by us

> Would it help to have it as a submodule? Not a fan of the approach, but if it's better for you shrug.

Don't see how this would help. Submodules are not a way to deal with naming issues. My overall opinion is that extracting is a bit pointless here. But again, it's just an opinion.

> confused The name is pretty self-explanatory to me, at most I can understand one wouldn't be sure what normalization means in this case, you would have that pain before. I'm more than open to better suggestions for the name though smile.

Without the benefit of hindsight, `normalizePathForRegExp` tells me nothing about what exactly this is doing. I have no idea why a path would need to be normalized for a regex, or what that normalization is. "normalize", in this particular case, falls under the same category as the "SomethingManager" classes we've learned to avoid

naps62 (approved):

PR #379, opened 2020-09-04
Fix SEO title

pfac:
Why:

* #372 refactored the SEO component and introduced an issue where the
  title of the homepage contains the separator that should only be
  visible when there's a page-specific title.

* Extracting the title building logic to its own function in the SEO
  component.

Before:
Screenshot from 2020-09-04 10-37-30

After:
Screenshot from 2020-09-04 10-37-44


jfranciscosousa (approved):
<3

PR #381, opened 2020-09-04
Fixes the intros of all old blog posts

pfac:
Why:

* Many of them still contain Markdown syntax or HTML tags, which are no
  longer rendered.


jfranciscosousa (approved):
I will not review every single intro. Just ship it

PR #387, opened 2020-09-16
Allow multiline text on the post SEO image generator

frm:
Previews:

out

out2

In both scenarios, the text is aligned to the bottom (meaning that if you put them side by side, the bottom line should match)

### DIFF
--- a/bin/generate-post-meta-image
+++ b/bin/generate-post-meta-image
@@ -1,34 +1,64 @@
 #!/bin/sh -e
+#
+# Example:
+#
+# BG="#222222" bin/generate-multiline-post-meta-image "First line" "Second line" out.png
 
-text="$1"
-bg="$2"
-out=$3
 res=2160x1080
 
 tmp=$(mktemp -d)
 
-if [[ "$bg" == "" ]]; then
-  bg="#2421AB"
-  convert -size $res xc:$bg $tmp/bg.png
+# generate background
+if [[ "$BG" == "" ]]; then
+  BG="#2421AB"
+  convert -size $res xc:$BG $tmp/bg.png
 else
   convert $2 -resize $res^ $tmp/bg-uncropped.png
   convert $tmp/bg-uncropped.png -gravity center -crop $res+0+0 -fill black -colorize 40% $tmp/bg.png
 fi
 
+# generate header
 convert -size $res xc:"#00000000" \
   -pointsize 100 -fill white -draw 'text 100,160 "Blog Post" ' \
   $tmp/text-top.png
 
-convert -size $res xc:"#00000000" \
-  -font ./src/fonts/acta-headline-extra-bold.woff \
-  -pointsize 165 -fill white -draw "text 100,925 '$text' " \
-  $tmp/text-bottom.png
+if [[ "$#" -eq "2" ]]; then
+  files="$tmp/text-top-line.png"
+  top_line="$1"
+  out="$2"
 
-convert \
-  -page +0+0 $tmp/bg.png \
-  -page +0+0 static/script-assets/blog-post-top.png \
-  -page +0+0 $tmp/text-bottom.png \
-  -layers merge +repage $out
+  # generate single lined text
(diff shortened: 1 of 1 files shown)


bin/generate-post-meta-image
```diff
@@ -1,34 +1,64 @@
...
 
 tmp=$(mktemp -d)
 
-if [[ "$bg" == "" ]]; then
-  bg="#2421AB"
-  convert -size $res xc:$bg $tmp/bg.png
+# generate background
+if [[ "$BG" == "" ]]; then
+  BG="#2421AB"
+  convert -size $res xc:$BG $tmp/bg.png
```
naps62, line 14: wait, something's wrong here
I don't understand why, it also seems wrong before your changes, but `$bg` was suposed to be not necessarily a color, but an image file

I don't see how the `[[ "$bg" == "" ]] makes sense (on either versions)

frm: @pfac weigh in on this?

I just moved `bg` from a script argument to an environment variable to avoid having to deal with script flags, opt parsing and variable number of arguments. Didn't change anything else.

pfac: I feel like I don't have much context here, the script was not exactly super clear to begin with, but this change is very odd to me.

First, it introduces a `BG` variable, which, if empty, will fallback to a default value and be used to create a background image. If it is not empty, it won't be used at all. Why is this a parameter at all? :confused: 

Second, I disagree with using environment variables instead of parameters. Positional parameters are given straight away to the script, and reading flags is exceptionally simple with `getopts`. Having to define environment variables in a command call is not a conventionally good experience for the user, and exposes the script to unintended outside interference if the user forgets to explicitly set the environment variable. In my opinion, the best approach would be with a flag through `getopts`; the second best would be a positional argument.

frm: > First, it introduces a BG variable, which, if empty, will fallback to a default value and be used to create a background image. If it is not empty, it won't be used at all. Why is this a parameter at all? 😕

Good question, the original scripts was already doing that.

Second: it's not exceptionally simple, it's difficult to adapt to other shells and I've been successfully in many scripts more recently, with this new approach. Using environment variables is a lot more intuitive, portable and easy to handle. And you don't need the boilerplate code for defining and parsing flags.

As for outside interference, not if you namespace it.

I can change it if you want, since this is not a common use case, but I don't agree with the reasoning you listed. I've been using it and it's so much better than flags. Script code is very much reduced, a lot more portable, a lot easier to remember when compared to flags, a lot more explicit. It's a change I've been glad to make, personally. Of course, this is personal experience and opinion, so I'm willing to be democratic in this.

pfac: I've been using `getopts` with `/bin/sh` for quite some time in a variety of systems. So far with no problems. There's also the older `getopt`, not as good looking but still very reliable from my experience.

For me both options have been equally portable, but yes, there is the possibility of a system not having `getopts` or `getopt`. As for being more intuitive, I disagree, using environment variables as one-time arguments is not more intuitive for me. Nor easier to handle, since it can be defined in the environment without the user realizing it. Namespacing does not prevent this at all.

I guess we'll have to agree to disagree. I'm just stepping in because you asked me, this is my opinion. If you don't agree with my reasonings, feel to free to proceed, there's no reason for me to block it. This discussion has little to no impact in the website as a whole.

pfac:
Is this blocked?

naps62:
@pfac kind of. I believe my initial question ended up unanswered and the discussion side-tracked heavity to bash implementation details

looking at the script, I don't understand how this could works (if it even works at all). Then again, it doesn't seem to be a problem introduced by this commit, but something that was already wrong. But merging changes to a script that's already broken seems kind of pointless anyway

I'll have to review this script on my own end (sorry, could have done this last friday, but missed this notification. I'm still getting coverflex spam on github :/ need to solve that)

naps62:
@frm I think this is now superseded by #405

frm:
Yep, let's close this.

PR #393, opened 2020-10-19
Upgrade IntersectionObserver polyfill

pfac:


gabrielpoca (approved):
I'm approving, but what do you about dropping this? https://caniuse.com/intersectionobserver

pfac:
> I'm approving, but what do you about dropping this? https://caniuse.com/intersectionobserver

Haven't considered it lately actually. We added this specifically for Safari at the time (even Gatsby lazy-loading images was failing because of this). I'll take a look and analyse whether we still need this. :+1:

PR #397, opened 2020-10-19
Refactor posts list component to use PascalCase

pfac:
Why:

* Following up on the work started in #362, but with a more limited
  scope to prevent breaking in progress work.

* Renaming the posts list component files to follow the Airbnb
  convention of naming the files after the same entity they export by
  default, which imposes using PascalCase for component files.


pfac:
Taking the executive decision of merging this. The code has been here for so long I no longer knew about it, so I just revised it myself.

PR #404, opened 2020-10-26
Refactor generated CSS module class names

pfac:
Why:

* By default Gatsby uses only the name of the module it gets from the
  filename, which means that most of our CSS modules become
  `index-something-XXXXX` when nesting all the files related to a
  component in a single directory (due to CSS module files being named
  `index.module.(s)css`.

* Tweaking `localIdentName` in the CSS Webpack loader so it uses the
  folder in the generated class name, by prepending the folder
  interpolation to the GatsbyJS default value for `localIdentName`
  (copied from [the source])

[the source]: https://github.com/gatsbyjs/gatsby/blob/18c182a574475c9c06dbbeb7ddae27f5cb008fe9/packages/gatsby/src/utils/webpack-utils.ts#L216

### DIFF
--- a/config/gatsby/plugins/sass.js
+++ b/config/gatsby/plugins/sass.js
@@ -1,0 +1,27 @@
+const path = require("path")
+const postCssUrl = require("postcss-url")
+const sass = require("sass")
+
+module.exports = (root) => [
+  {
+    resolve: "gatsby-plugin-sass",
+    options: {
+      cssLoaderOptions: {
+        localIdentName: "[folder]-[name]--[local]--[hash:base64:5]",
+      },
+      implementation: sass,
+      includePaths: [
+        path.resolve(root, "node_modules"),
+        path.resolve(root, "src"),
+      ],
+      postCssPlugins: [
+        postCssUrl([
+          {
+            filter: "**/fonts/inline/*",
+            url: "inline",
+          },
+        ]),
+      ],
+    },
+  },
+]
--- a/config/gatsby/plugins.js
+++ b/config/gatsby/plugins.js
@@ -1,26 +1,13 @@
 const path = require("path")
-const postCssUrl = require("postcss-url")
-const sass = require("sass")
 
 const ROOT = path.resolve(__dirname, "../..")
 
+const sassConfig = require("./plugins/sass")(ROOT)
 const rssFeedConfig = require("./plugins/rssFeed")(ROOT)
 const manifestConfig = require("./plugins/manifest")(ROOT)
 
 module.exports = [
-  {
-    resolve: "gatsby-plugin-sass",
-    options: {
-      implementation: sass,
-      includePaths: [
-        path.resolve(ROOT, "node_modules"),
-        path.resolve(ROOT, "src"),
-      ],
-      postCssPlugins: [
-        postCssUrl([{ filter: "**/fonts/inline/*", url: "inline" }]),
-      ],
-    },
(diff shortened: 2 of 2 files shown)


config/gatsby/plugins/sass.js
```diff
@@ -0,0 +1,27 @@
+const path = require("path")
+const postCssUrl = require("postcss-url")
+const sass = require("sass")
+
+module.exports = (root) => [
+  {
+    resolve: "gatsby-plugin-sass",
+    options: {
+      cssLoaderOptions: {
+        localIdentName: "[folder]-[name]--[local]--[hash:base64:5]",
```
gabrielpoca, line 10: I'm confused, is this dev only? Because in prod you only need the `hash` part.

gabrielpoca (commented):
I'm still confused about why we are using both sass and postcss

config/gatsby/plugins/sass.js
pfac: This is used in both places. GatsbyJS doesn't seem to make any distinction between the two in the CSS Loader configuration, probably due to the fact the full hash produces bigger classes than this setup (https://github.com/webpack-contrib/css-loader/issues/406).

Since GatsbyJS doesn't make the distinction I didn't bother with it, but we distinguish it ourselves. Would you prefer to use `[hash:base64:20]` in production? Or any other hash length?

pfac:
> I'm still confused about why we are using both sass and postcss

Because there are advantages and shortcomings to both of them, and they work together well enough we can pick and choose. We started by using SASS for the nesting (#262), then hen for `include-media` (#293), and for the compile-time variables (#296, #325). We still use PostCSS for CSS modules and other stuff, mostly as it comes with GatsbyJS by default.

pfac:
@gabrielpoca ping on this. Answered all your comments. :pray:

gabrielpoca:
@pfac I thought that I had approved this PR. My comments are not blocking. Regarding the hash, I think it's a good thing to avoid shipping unnecessary bytes, so `[hash:base64:20]` is likely more than enough.

Regarding SASS, Gatsby comes with a lot of things by default, that's one of its advantages, but I don't think we should be picking up everything. It's fine to use SASS, but since you are already using PostCSS, there's no reason to use it. The things you need are also available in PostCSS. Using one tool is easier in my opinion, special when they have a huge overlap like these two.

gabrielpoca (approved):

pfac:
> @pfac I thought that I had approved this PR. My comments are not blocking. Regarding the hash, I think it's a good thing to avoid shipping unnecessary bytes, so `[hash:base64:20]` is likely more than enough.
> 
> Regarding SASS, Gatsby comes with a lot of things by default, that's one of its advantages, but I don't think we should be picking up everything. It's fine to use SASS, but since you are already using PostCSS, there's no reason to use it. The things you need are also available in PostCSS. Using one tool is easier in my opinion, special when they have a huge overlap like these two.

Sorry I seem to have missed your comment and merged away. Will hotfix the hash to `[hash:base64:20]` in production. :+1: 

As for PostCSS, while the things we need also exist in PostCSS they might work the same way or work as well as in SASS. When we added SASS that was the case, we were spending too much time fighting PostCSS plugins for things SASS already give us for free with no impact in the final bundle.

I don't mind if anyone wants to spend some time making the required configurations so we can drop SASS, but I won't do it myself since I don't see it as a problem.

PR #405, opened 2020-10-29
Refactors the banner generator into a sensible thing

naps62:
Finally got some time to look into this

The previous script was a bit of a mess. It "worked", but it was hard to work with once we needed to improve it

As planned, I'm now refactoring it into something we can actually look at without crying.

* adds positional arguments, solving a whole host of problems
* uses `-annotate` instead of `-draw` which allows for multiline text, and better positioning support
* merges most imagemagick calls into just a couple of them

I believe #387 is now no longer needed. In my view, it attempted to solve hacks by adding additional hacks (e.g.: using environment variables for what should be arguments)

### DIFF
--- a/bin/generate-post-meta-image
+++ b/bin/generate-post-meta-image
@@ -4,7 +4,7 @@ tmp=$(mktemp -d)
 bg="#2421AB"
 res=2160x1080
 
-while [[ $# -gt 0 ]] && [[ "$1" == "--"* ]]; do
+while [[ $# -gt 0 ]] && [[ "$1" == "--"* || "$1" == "-"* ]]; do
   case "$1" in
     "--resolution" | "-r")
       res=$2
@@ -18,7 +18,7 @@ while [[ $# -gt 0 ]] && [[ "$1" == "--"* ]]; do
       shift
       ;;
 
-    "--bg" | "-b")
+    "--background" | "-b")
       bg=$2
       shift
       shift
--- a/bin/generate-post-meta-image
+++ b/bin/generate-post-meta-image
@@ -1,34 +1,68 @@
 #!/bin/sh -e
 
-text="$1"
-bg="$2"
-out=$3
+tmp=$(mktemp -d)
+bg="#2421AB"
 res=2160x1080
 
-tmp=$(mktemp -d)
+while [[ $# -gt 0 ]] && [[ "$1" == "--"* ]]; do
+  case "$1" in
+    "--resolution" | "-r")
+      res=$2
+      shift
+      shift
+      ;;
 
-if [[ "$bg" == "" ]]; then
-  bg="#2421AB"
-  convert -size $res xc:$bg $tmp/bg.png
+    "--text" | "-t")
+      text=$2
+      shift
+      shift
+      ;;
+
+    "--bg" | "-b")
+      bg=$2
+      shift
+      shift
+      ;;
+  esac
+done
+
+out=$1
+
+
+if [[ -f "$bg" ]]; then
+  # bg is a file. scale & crop the image to intended dimensions,\
+  # and add a transparent layer on top
+  convert $bg \
+    -resize $res^ \
+    -gravity center \
+    -crop $res+0+0 \
+    -fill black \
+    -colorize 40% \
+    $tmp/bg.png
 else
-  convert $2 -resize $res^ $tmp/bg-uncropped.png
-  convert $tmp/bg-uncropped.png -gravity center -crop $res+0+0 -fill black -colorize 40% $tmp/bg.png
+  # bg will be assumed to be a color, used as is
+  convert -size $res \
+    xc:$bg \
+    $tmp/bg.png
 fi
 
-convert -size $res xc:"#00000000" \
-  -pointsize 100 -fill white -draw 'text 100,160 "Blog Post" ' \
-  $tmp/text-top.png
-
-convert -size $res xc:"#00000000" \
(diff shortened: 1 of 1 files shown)


bin/generate-post-meta-image
```diff
@@ -1,34 +1,68 @@
...
 
-text="$1"
-bg="$2"
-out=$3
+tmp=$(mktemp -d)
+bg="#2421AB"
 res=2160x1080
 
-tmp=$(mktemp -d)
+while [[ $# -gt 0 ]] && [[ "$1" == "--"* ]]; do
```
pfac, line 7: Either I'm missing something or these conditions do not allow for short flags. Does "-r" match here for instance?

bin/generate-post-meta-image
```diff
@@ -1,34 +1,68 @@
...
+    "--text" | "-t")
+      text=$2
+      shift
+      shift
+      ;;
 
-convert -size $res xc:"#00000000" \
-  -pointsize 100 -fill white -draw 'text 100,160 "Blog Post" ' \
-  $tmp/text-top.png
+    "--bg" | "-b")
```
pfac, line 21: Shouldn't the long version be `background`?

bin/generate-post-meta-image
```diff
@@ -1,34 +1,68 @@
...
+  -gravity northwest \
+  -annotate +100+160 'Blog Post' \
+  -gravity southwest \
+  -font ./src/fonts/acta-headline-extra-bold.woff \
+  -annotate +100+100 "$text" \
+  -layers merge \
+  +repage \
+  $out
 
-# Optimize image for smaller footprint
```
pfac, line 33: Why remove this comment? Seems to be consistent with the rest of the script, where there's a comment explaining what's intended with each "section".

pfac (commented):
Assuming the manual flag handling is to be able to use long flags, I personally don't get the value when there's a built-in way to handle just the short flags more simply and there are just a couple of them. Seems like over-complicating things to me.

Nonetheless, if we're adding flags I think this is missing a usage message, which should be shown with `--help`/`-h`, and with invalid flags. One of the problems with just ignoring these invalid flags is that their values will be considered positional arguments. In fact with this manual approach, if the user mistypes the flag, the script will stop looking for flags and consider the value of the bad flag the output file name.

bin/generate-post-meta-image
naps62: true. missed that case (this was originally copy/pasted from another script

but in hindsight, and answering your other comment, I guess I can just remove support for the shorthand syntax

bin/generate-post-meta-image
naps62: because to be honest, the command itself is pretty much self-documenting

naps62:
@pfac I agree that a `--help` option would be helpful, but that's out of scope of what I wanted to do here. the goal was to make the script easier to work with. that is an extra feature, that can easily be added later if someone really wants it

bin/generate-post-meta-image
pfac: Not sure about it if someone doesn't know what Bundler or `image_optim` are, but if you feel it is I'm ok with removing.

pfac:
> @pfac I agree that a `--help` option would be helpful, but that's out of scope of what I wanted to do here. the goal was to make the script easier to work with. that is an extra feature, that can easily be added later if someone really wants it

I'm usually the first one saying something is out of the scope of a PR, but I don't agree with you here. `-h/--help` makes the script easier to work with by providing information on how it is supposed to be used (and what it does). Code-wise, it also makes sense since you're adding support for flags in this PR.

> (from [your comment](https://github.com/subvisual/subvisual.com/pull/405#discussion_r514400214)) but in hindsight, and answering your other comment, I guess I can just remove support for the shorthand syntax

I don't get this conclusion. Support for the shorthand syntax is what is builtin with things like `getopts`. To me it seems the simplest approach would be to support only the shorthand syntax, not the other way around.

**Anyhow**, I don't feel I have any authority to tell you to do A or B, so I won't press any more on either of these things. I'll leave it up to you.

naps62:
> I'm usually the first one saying something is out of the scope of a PR, but I don't agree with you here. -h/--help makes the script easier to work with by providing information on how it is supposed to be used (and what it does). Code-wise, it also makes sense since you're adding support for flags in this PR.

I'm honestly failing seeing how this is an answer to my initial argument.
This PR tries to does one thing: makes the script easier to work with
You proposed an additional thing: create a help argument. Which makes sense here, just as it would have made sense in the previous version. It's a helpful thing, of course, but was not the goal of this PR, nor does this change require that in any way

Same thing for `getopt`. I did this change in the way I knew how, and in order to make the best I could without spending too much time.
I don't really know how to work with `getopt`, and have always used a case statement similar to this one for dealing with positional args (and at this point, I just copy & paste it from one script to the next). Between going to the trouble of learning `getopt`  and implementing it here, versus discarding something that is not even necessary in the first place, I went with the easy choice

pfac:
As I said I'll leave it up to you. Just wanted to help.

Let me know if you're going to do any change regarding my two unresolved comments, otherwise I can approve you right away.

bin/generate-post-meta-image
naps62: fixed

bin/generate-post-meta-image
naps62: indeed. fixed

naps62:
@pfac both comments fixed. shall we get on with this?

pfac (approved):

PR #414, opened 2020-12-09
Fixes a few h1 tags that should be h2

naps62:


pfac:
Do we really have to remove the title from the posts that have only one and at the top? That's supposedly not rendered at all, so shouldn't make a difference as far as I'm aware. That's why I left them, at least in a Markdown previewer you get to see it since frontmatter is ignored.

naps62:
@pfac sorry, I distinctively remembering answering here, but I guess I forgot to submit somehow

To answer your comment, no, at no point I said we needed to remove the title. I just happened to do so as a bi-product of what I was doing (using `ripgrep` to search for posts that wrongfully used h1 instead of h2)

pfac:
> @pfac sorry, I distinctively remembering answering here, but I guess I forgot to submit somehow
> 
> To answer your comment, no, at no point I said we needed to remove the title. I just happened to do so as a bi-product of what I was doing (using `ripgrep` to search for posts that wrongfully used h1 instead of h2)

I'm very confused. Not sure I understand what you mean, or that you understood what I meant, or both.

I didn't say you said (:confused:) we needed to remove the title. In some posts, for instance `src/posts/2020-06-04-how-we-set-and-track-goals/index.md`, this change is removing the `h1` in the Markdown. But these `h1` tags are not rendered in the post page (styled with `display:none`). Do we actually have to remove them, i.e., is there any problem and they shouldn't be there at all in the markup?

naps62:
You asked

> Do we really have to remove the title 

What I'm saying was that no, we don't, but that ended up being done anyway as a bi-product of my search/replace. I'm perfectly aware we don't need to remove them, but I also don't see a strong reason to keeping them when only 2 posts had them

pfac:
> What I'm saying was that no, we don't, but that ended up being done anyway as a bi-product of my search/replace. I'm perfectly aware we don't need to remove them, but I also don't see a strong reason to keeping them when only 2 posts had them

I don't understand your reasoning, sorry. "We don't have to change this but my tool did it anyway, but it's only 2 posts so there's no reason to keep them either". I gave you a reason:

> at least in a Markdown previewer you get to see it since frontmatter is ignored.

Still, it's a useless discussion, I was just baffled by the reasoning. Please proceed.

pfac (approved):

PR #415, opened 2020-12-09
Adds WebP support to both sharp and remark image loaders

naps62:
Because modern image compression.

Before (only our homepage hero):
image

After:
image


gabrielpoca (approved):
Nice 👍 
I thought Firefox didn't support webp yet.

PR #416, opened 2020-12-09
Pre-optimizes images. Saves a bunch of space

naps62:
I'll need input here, because I'm not sure I missed something

By looking at our code, the maximum width we use in our images is `maxWidth: 960px` (the hero images)
Which means that, accounting for retina, we need raw images to be at most 1820

Or are we considering 3x/4x retina? Don't even know if that's a thing

What I did here was copy the script [recommended by gatsby](https://www.gatsbyjs.com/docs/preoptimizing-images/) and run it on all our images.

Let me know if this isn't viable though


pfac:
I'm perfectly fine with this in concept, but since it is using lossy compression, can you run the preview by one of the designers on Slack first please? Just so they don't back at us yelling. :pray: 

Also, while there's no need to do it right away, we should then add a check to the CI that runs this and fails if there are any changes, to prevent merging non optimized images.

naps62:
that one is harder, because optimizers might in some generate changes even if tiny ones. A CI rule would potentially need to measure the before & after sizes of each image, check that against an acceptable threshold, and fail if any image exceeds it

sounds interesting, but definitely not trivial

pfac:
I was actually thinking of something way simpler:
1. Checkout repo
2. `bin/preoptimize-images`
3. `git status` must be blank.

That way whatever `gatsby build` may do will not affect this.

naps62:
But that's what I meant. Step 3 isnt' necessarily true. optimizers will often end up change a few bytes in an image (not sure if there's timestamps involved in some cases, or dependant on the version used)

It would mean you were forced to use the exact same optimizer used by the server, which is a bit too restrictive in my opinion

naps62:
also FYI, Francisco already reviewed this preview. it's good 2 go

pfac (approved):

PR #417, opened 2020-12-09
`convert -strip -interlace Plane -quality 80` all the JPGs

naps62:
AKA: convert all JPGs to progressive, stripped of comments, and 80% quality

This shouldn't affect website performance too much, because gatsby already processes images,
but it should definitely help with repo size and build times


pfac:
I'm not comfortable with lossless compression without a designer's approval. Can you run the preview by one of them on Slack first? :pray:

naps62:
fine, but you do realise that everytime you use `image_optim`, you're also using lossy compression, right?

I can run this through them, but I honestly don't think this should be a norm. 80% quality jpegs are pretty much a norm, and the impact in quality is small enough for this to be considered a technical decision rather than a design one

lossy compression on PNGs, now that would be a different scenario, but jpegs are tipically used for photos and other kinds of images where such losses are perfectly acceptable

pfac:
> fine, but you do realise that everytime you use `image_optim`, you're also using lossy compression, right?
:raised_eyebrow: You sure? According to their [README](https://github.com/toy/image_optim) it is always lossless by default, the lossy compression is optional.

> I can run this through them, but I honestly don't think this should be a norm. 80% quality jpegs are pretty much a norm, and the impact in quality is small enough for this to be considered a technical decision rather than a design one
> 
> lossy compression on PNGs, now that would be a different scenario, but jpegs are tipically used for photos and other kinds of images where such losses are perfectly acceptable

I'm not arguing with the value of your proposition, just tired of hearing smack about how developers don't care or whatever (pun intended). I just don't want to hear them complain afterwards, doesn't hurt to include them in the decision.

gabrielpoca:
I wouldn't do this for one reason: I like to have the image at its highest resolution so I can visually compare it against the final processed version. Still, I don't think this change affects. I also don't agree that this a design decision, website performance, and image optimization is clearly the responsibility of a developer until it affects the final result.

pfac:
> until it affects the final result.

:arrow_up: That is precisely my point. It does not hurt to include them in judging whether the final result is being affected. Unless you have an objective way of defining "being affected", it is subjective and therefore I believe they should be included.

**In any case**, if you both are against including the designers, they by all means please disregard my comment and proceed without my input.

naps62:
I am not against, especially because this was a very wide change across all images
anyway, francisco already gave his :+1: on this one as well

pfac (approved):

naps62:
Sorry, missed a couple of comments here:

> 0You sure? According to their README it is always lossless by default, the lossy compression is optional.

Guess I was wrong, sorry about that. Didn't notice that part of the readme, and I had always been under the assumption that the defaults included lossless as well.

@gabrielpoca I'm not sure I understand your point
if the website only uses images with 960px width (1820 with retina), then why would you want to keep a 3000px image? can't the same visual comparison be done if you scale it down to 1820?

gabrielpoca:
It can, but I'm also 100% sure that I'll forget to ask for a bigger image if those sizes increase for some reason.

gabrielpoca:
@naps62 can we close this?

naps62:
you mean merge or cancel?

gabrielpoca:
@naps62 I would rather cancel for the reason I mentioned: if the design changes and we need bigger images we'll have a berbicacho.

naps62:
@gabrielpoca I'm not following.

The only change here was to the jpeg params for each image (using progressive jpegs with 80% quality, which is pretty much the standard). There is no change to image resolution

"if the design changes and we need bigger images", then we'd still have that problem.
This is purely saving space

gabrielpoca:
@naps62 for some reason I thought this would change the resolution. In that case, let's merge it. Do you want to re-run the thing for new images?

gabrielpoca (approved):

gabrielpoca:
@naps62 bump

PR #419, opened 2020-12-09
Updates minor versions of several packages

naps62:
* All changelogs were checked for breaking changes
* Local copy booted locally to test everything still worked
* Testing a couple of times, saw a decreate from ~25seconds to ~15seconds on the image processing step. So I expect this to shave a few seconds off our build as well

I'll probaby try to update major versions of a few packages as well, but on a later PR
code review


pfac (approved):
I suggest doing the major upgrades package-by-package instead of a single PR, to make it easy to understand which one if any upgrade causes issues after merging.

PR #420, opened 2020-12-09
It seems we're using dart-sass, and not node-sass

naps62:


pfac (approved):
Can confirm, we are. Shame on me for forgetting to remove `node-sass`.

PR #422, opened 2020-12-14
More updates

naps62:
Even though some of the upgrades here are major versions, acording to
their changelogs, there are no breaking changes relevant to us. In fact
the only one with breaking changes was framer-motion.


naps62:
@pfac I think most of the other ones aren't upgradeable right now, because they all specify react@^16.x.x as a dependency. So untill gatsby itself doesn't support react 17, all the others are locked in place as well

pfac (approved):

PR #438, opened 2021-04-23
Announcing Still

gabrielpoca:
[Docs for feedback](https://docs.google.com/document/d/13jMa9mktWJc1Wh1sZru7P3R5ZFd1-abtTdqKJm-Hhwc/edit#).


gabrielpoca:
@pfac can we make the image's width smaller? This doesn't look right.

pfac:
You can if you use custom HTML. By default it will always be forced to 100%.

gabrielpoca:
Will custom html also import the image?

gabrielpoca:
I tried it and it breaks because gatsby-image does some magic underneath

pfac:
Sorry, forgot about that. Easiest way is not to fight it, just go on Figma or something and put your image inside a bigger container.

The custom HTML way, you add the image, hide it with css, and manually copy the src into a new tag. I don't recommend it.

gabrielpoca:
The bigger container solution will make it look terrible on mobile :/

pfac:
I know but I don't have a better solution :/

I don't think it's worth the effort at the moment, but there are two possible solutions for this long-term:
1. Revisit the image styles for the blog post content. I've tried before to make it more "universal" but I couldn't figure out a way. Feel free to give it a try yourself though.
2. Use MDX and have posts use React components. This is by far my favorite option but never felt it was worth the effort.

pfac (approved):

PR #457, opened 2021-09-22
Only show announcement every 30 days

gabrielpoca:


gabrielpoca:
I asked on Slack and no one seems to be against it. I'll wait a bit longer to see if someone says anything.

naps62 (approved):

PR #461, opened 2021-10-17
Auto generate an SEO image for blog posts that don't have any

gabrielpoca:


gabrielpoca:
No, it's not in there, it doesn't work without this line https://github.com/subvisual/subvisual.com/pull/461/files#diff-f9ba7ecf60ea5af1dae0abe92b1909c94ace3aa290b376f41b35573f1d02203fR6

PR #487, opened 2023-02-17
Replace gtag with GoatCounter

gabrielpoca:
Extra: I had to upgrade the node version because it was breaking my whole system and it didn't make any sense to use such an old version of node.


mirhamasala (approved):
🌈

PR #492, opened 2023-02-24
Improve Netlify CMS

gabrielpoca:
* Adds preview using the blog post's page components.
* Adds editorial workflow.

I had to update the way the post's path worked to make the editorial workflow work. It's a long story, but we can't have "/" in the post's path.

### DIFF
--- a/config/gatsby/plugins/rssFeed.js
+++ b/config/gatsby/plugins/rssFeed.js
@@ -45,7 +45,7 @@ module.exports = () => [
                 description: intro,
                 guid: url,
                 title,
-                url: url,
+                url,
               }
             }),
         },
--- a/config/gatsby/plugins/rssFeed.js
+++ b/config/gatsby/plugins/rssFeed.js
@@ -14,7 +14,7 @@ module.exports = () => [
               ) {
                 nodes {
                   fields {
-                    slug
+                   url
                   },
                   frontmatter {
                     author {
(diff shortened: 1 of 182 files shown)


config/gatsby/plugins/rssFeed.js
```diff
@@ -43,9 +43,9 @@ module.exports = () => [
                 ],
                 date,
                 description: intro,
-                guid: slug,
+                guid: url,
                 title,
-                url: slug,
+                url: url,
```
mirhamasala, line 48:
```suggestion
                url,
```

src/components/BlogPostPreview.js
```diff
@@ -0,0 +1,60 @@
...
+
+import * as styles from "~/src/templates/blog/post.module.scss"
+import * as titleStyles from "~/src/components/blog/post/header.module.css"
+
+const BlogPostPreview = (props) => {
+  const md = props.entry.getIn(["data", "body"])
+  const [html, setHTML] = useState("")
+
+  useEffect(() => {
+    ;(async () => {
```
mirhamasala, line 22: Is this semicolon on purpose?

mirhamasala (approved):

src/components/BlogPostPreview.js
```diff
@@ -0,0 +1,60 @@
...
+
+        setHTML(value)
+      } catch (e) {
+        console.error(e)
+      }
+    })()
+  }, [md])
+
+  return (
+    <div className={styles.root} style={{ paddingTop: "10px" }}>
```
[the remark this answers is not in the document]
gabrielpoca: I'm just trying to override the existing styles and not change any CSS. This is just the preview.

src/components/BlogPostPreview.js
gabrielpoca: Prettier adds it automatically. I think it's to avoid some syntax error because there's a ( after {.

PR #507, opened 2023-03-24
Add category to posts in Netlify CMS

gabrielpoca:


jreis13 (changes requested):
'Personal Growth' was one suggested category that is missing on the options here

gabrielpoca:
@jreis13 I didn't add it because that's more of a strategic decision. I know there was a comment for that, but I would leave the decision to Laura and Júlia.

jreis13 (approved):

PR #521, opened 2023-04-04
Show categories on the index page

gabrielpoca:
In this PR:
* update the schema to allow querying the categories for each post.
* change the posts' layout to two columns.
* show the categories for each post.
* each category starts with rounded corners, and there's a probability of 30% of those rounded corners becoming square.


src/components/Categories/index.module.scss
```diff
@@ -0,0 +1,21 @@
...
+  margin-right: 12px;
+  height: 32px;
+  font-size: 12px;
+  display: flex;
+  align-items: center;
+
+  &:nth-child(2n) {
+    color: var(--c-purple-1);
+  }
+}
```
mirhamasala, line 21: Are we not using style lint to order CSS properties?

src/components/PostList/index.jsx
```diff
@@ -2,14 +2,20 @@ import React from "react"
 import PropTypes from "prop-types"
 import dateFormat from "dateformat"
 import { Link } from "gatsby"
+// import { GatsbyImage, getImage } from "gatsby-plugin-image"
```
mirhamasala, line 5: Leftover

src/components/PostList/index.jsx
```diff
@@ -23,19 +29,38 @@ function Entry({ author, date, intro, path, title }) {
...
+            alt=""
+            image={getImage(author.photo.vertical)}
+            style={{
+              marginRight: 16,
+              height: 40,
+              width: 40,
+              borderRadius: "50%",
+            }}
+          />
+        )} */}
```
mirhamasala, line 43: Leftover?

src/components/PostList/index.module.scss
```diff
@@ -2,10 +2,17 @@
 
 .posts {
   width: 100%;
-  max-width: 980px;
+  max-width: 1039px;
```
mirhamasala, line 5: That's specific

src/components/PostList/index.module.scss
```diff
@@ -51,22 +39,11 @@
...
-    margin-bottom: 0;
-  }
-
-  @include media("grid-10") {
-    grid-column-start: 2;
-  }
-
   /**
    * Fix for Firefox font rendering issues.
    */
```
mirhamasala, line 49: Is this still needed?

src/components/Categories/index.module.scss
gabrielpoca: ahaha didn't you remove that? I think I just don't have it configured on my editor. I'll take care of that 👍

src/components/PostList/index.module.scss
gabrielpoca: Yeah, the designers use a 13 columns grid but only use 11 columns. Either way, it's just a weird/random number :/

src/components/PostList/index.module.scss
gabrielpoca: I guess? I'm not sure if something has changed since this was added.

src/components/Categories/index.module.scss
mirhamasala: No, in the end I didn't remove anything ..

gabrielpoca: 👍

src/components/PostList/index.module.scss
```diff
@@ -76,6 +76,21 @@
...
+  width: 40px;
+  border-radius: 50%;
+}
+
+.noPhoto {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  color: var(--c-white-1);
+}
```
mirhamasala, line 92: Could you add all these styles under `.avatar`? I'm wondering if you need two classes. A default/faalback `background-color` would be nice too.

src/components/PostList/index.jsx
```diff
@@ -29,17 +43,22 @@ function Entry({ author, date, intro, path, title, categories }) {
         </Link>
       </p>
       <div className={styles.info}>
```
mirhamasala, line 45: I think we should only show this part if `author` exists. Check line 64.

config/node/createSchemaCustomization.js
```diff
@@ -45,12 +29,45 @@ module.exports = ({ actions }) => {
...
+        tags: [String]!,
+        title: String!,
+        seoDescription: String,
+        seoImage: String,
+      }
+
+      type CategoryYaml implements Category & Node @dontInfer {
+        key: String!,
+        label: String!  
+        initials: String,
```
gabrielpoca, line 53: Why do we need the `initials` in the categories?

src/components/Header/index.jsx
```diff
@@ -9,10 +10,11 @@ import NavMenu from "../NavMenu"
 import * as styles from "./index.module.scss"
 
 function Header() {
+  const hasJavascript = useDetectJavascript()
   const { scrollY, innerWidth } = useWindowState()
 
-  const isDesktop = innerWidth > 950
-  const isScrolled = scrollY > 756
+  const isDesktop = hasJavascript ? innerWidth > 950 : 0
+  const isScrolled = hasJavascript ? scrollY > 756 : 0
```
mirhamasala: This should be reverted

src/components/FooterGrid/index.module.scss
```diff
@@ -5,11 +5,17 @@
...
+  padding-left: 20px;
+
   @media (min-width: 768px) {
     justify-self: center;
     grid-row-gap: 56px;
     grid-template-columns: 1fr 1fr;
     max-width: 80%;
+
+    padding-right: 0;
+    padding-left: 0;
```
mirhamasala, line 18: I'm not sure if this is actually necessary. But if it is, you can have a look at `base.scss` and see what Gabx did for `--post-max-width` and `--post-body-max-width`

src/components/NewsletterCTA/index.module.scss
```diff
@@ -12,34 +12,37 @@
...
-  margin-bottom: 84px;
+  padding-right: 60px;
+
+  font-family: "colfax-web", sans-serif;
+  font-size: 56px;
+  font-weight: 400;
+  line-height: 56px;
 
   span {
     color: #ff7b9b;
```
mirhamasala, line 43: Let's add this to `base.scss`

PR #523, opened 2023-04-07
Add highlighted posts

gabrielpoca:


src/components/Author/index.module.scss
```diff
@@ -0,0 +1,6 @@
+.author,
+.date {
+  font-size: 16px;
+  line-height: 21px;
+  display: block;
```
mirhamasala, line 5: Stylelint?

src/components/HighlightedPosts/index.module.scss
```diff
@@ -0,0 +1,112 @@
...
+
+.image {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  max-width: 442px;
+  padding: 36px 16px 26px;
+
+  @include media(">=tablet") {
+    padding-top: 16px 16px 0 0;
```
mirhamasala, line 60: This can't be right

src/components/HighlightedPosts/index.module.scss
```diff
@@ -0,0 +1,112 @@
...
+  margin-bottom: 40px;
+
+  font-family: "Acta", serif;
+  font-size: 56px;
+  font-weight: 400;
+  line-height: 120%;
+}
+
+.dots {
+  display: flex !important;
```
mirhamasala, line 93: Instead of using `!important` I believe you should be able to use `:global` with CSS modules when wanting to override external library styles

src/components/SectionTitle/index.jsx
```diff
@@ -0,0 +1,7 @@
+import React from "react"
+
+import * as styles from "./index.module.scss"
+
+export default function HighlightedPosts({ children }) {
```
mirhamasala, line 5: Should be`SectionTitle`?

src/components/SectionTitle/index.module.scss
```diff
@@ -0,0 +1,19 @@
...
+  padding-left: 82px;
+
+  &:before {
+    content: "";
+    width: 62px;
+    position: absolute;
+    left: 0;
+    top: 50%;
+    height: 1px;
+    background-color: #9563ff;
```
mirhamasala, line 16: `--c-purple-1`?

src/pages/blog.module.scss
```diff
@@ -1,15 +1,14 @@
 @import "common/breakpoints";
 
 .root {
-  padding: 28px 20px 56px;
+  --x-padding: 20px;
```
mirhamasala, line 4: Why the abstraction?

src/components/HighlightedPosts/index.js
```diff
@@ -0,0 +1,114 @@
...
+          slidesToScroll: 1,
+          adaptiveHeight: true,
+          nextArrow: <NextArrow />,
+          prevArrow: <PrevArrow />,
+        }}
+        dotsClass={styles.dots}
+      >
+        {posts.map(
+          ({ path, intro, title, cover, categories, author, date }) => (
+            <div>
```
mirhamasala, line 84: I think you have an extra `div` here?

src/components/HighlightedPosts/index.module.scss
```diff
@@ -0,0 +1,112 @@
+@import "common/breakpoints";
+
+.root {
+  position: relative;
+
+  margin-bottom: 89px;
```
mirhamasala, line 6: I believe best practice is to move layout spacings to the parent / a layout component, etc.

gabrielpoca: Normally yes, and I was the person that came up with that rule in our context xD but I think we can safely ignore it because its purpose is to make components re-usable, and this one is 100% not going to be re-used

src/components/HighlightedPosts/index.js
gabrielpoca: Na, I need it because the library has conflicts with the next div that sets the grid.

src/components/HighlightedPosts/index.module.scss
gabrielpoca: Not in this case. The library itself allows me to specify a class to override its styles. I'm not supposed to reference its classes directly

src/pages/blog.module.scss
gabrielpoca: It was something more complex before 👍

src/components/HighlightedPosts/index.module.scss
mirhamasala: Sure. Though you're also removing the predicability of where spacings are set.

jreis13 (commented):
I'm not sure if that's the reason but it feels like whenever we have a post image in the highlighted article, we can't click on the author's name to navigate to his page

src/components/HighlightedPosts/index.js
```diff
@@ -0,0 +1,118 @@
...
+                <div className={styles.details}>
+                  <Link to={path}>
+                    <h3 className={styles.title}>{title}</h3>
+                  </Link>
+                  <span className={styles.intro}>{intro}</span>
+                  <div className={styles.author}>
+                    <Author author={author} date={date} />
+                  </div>
+                  {categories?.length && (
+                    <div className={styles.categories}>
```
jreis13, line 97: categories class isn't being used at the moment but the color should be blue

src/components/HighlightedPosts/index.module.scss
gabrielpoca: You think so? I can never guess where the margins are until I do "inspect element" and look at which component it is. I'm not against following the rules; it's just that the rules are not supposed to be applied blindly.

gabrielpoca:
@jreis13 addressed both 👍

jreis13 (approved):
LGTM 🚀

PR #524, opened 2023-04-07
Download all remote cover images

gabrielpoca:
This will allow us to remove a lot of custom code that has to distinguish between local images and remote images.


mirhamasala (approved):
Nice!

PR #530, opened 2023-05-03
Optimize images

gabrielpoca:
PNGs are huge and inefficient. In this change, I moved some images to use StaticImage that coverts the images in the best format supported by the browser. The footer I only changed to jpeg because it was not worth the effort of using StaticImage. I've kept the originals around in case we want to revise the quality.


jreis13 (approved):
Looks good 🚢

mirhamasala (approved):

PR #531, opened 2023-05-05
A bunch of small changes

gabrielpoca:
* Highlighted posts don't change height when navigating.
* Categories render the same shape every time and in every place.
* Filter by category.
* Save selected categories and loaded posts in the URL to allow navigating back and forth.
* Rotate images in highlighted posts in different directions.
* Fix missing pictures and initials on the authors' pages.
* Other small visual fixes.


jreis13 (approved):
LGTM 🚀

PR #551, opened 2023-05-23
Update team photos

### LINKED ISSUE #542: Change avatar photos
The avatars are blue at the moment, which is ironically right, but in this case, not.

jreis13:
Why:

- Closes #542 

These changes address the need by:

- Delete unused `horizontal` from Photo
- Delete horizontal versions of avatars
- Replace vertical for avatar
- Update team photos
- Add photos from other team members


config/node/createSchemaCustomization.js
```diff
@@ -18,8 +18,7 @@ module.exports = ({ actions }) => {
       }
 
       type Photo {
-        horizontal: File! @fileByRelativePath,
-        vertical: File! @fileByRelativePath,
+        avatar: File! @fileByRelativePath,
```
mirhamasala, line 21: Should we place this directly under the `Author` type? If there's just one photo anyway? CC: @gabrielpoca

gabrielpoca: Yeah, I think so. I don't see why we need this Photo type.

mirhamasala: @jreis13 Will you take care of this?

mirhamasala: When that's done, we can review again and merge this. Then, people can check their photo live, and request a change if they want.

mirhamasala (approved):

PR #553, opened 2023-05-25
Add related articles

### LINKED ISSUE #543: Add related articles
> [Design](https://www.figma.com/file/iZtsffS8kRYeIG5xrhbp1O/Subvisual-Blog-(Client)?type=design&node-id=1-2086&t=M4amWVpi3qHx3a43-4)

jreis13:
Why:
- Closes #543 

These changes address the need by:
- Add `PostRelatedPosts` component & styles
- Add a query to get categories
- Add `PostRelatedPosts` on Post template

Screenshot 2023-05-26 at 13 35 07


src/components/PostRelatedArticles/index.jsx
```diff
@@ -0,0 +1,71 @@
+import { graphql, useStaticQuery } from "gatsby"
+import React from "react"
+
+import * as styles from "./index.module.scss"
+
+export const query = graphql`
+  {
+    allMarkdownRemark(
+      filter: { fileAbsolutePath: { regex: "/src.posts/" } }
+      sort: [{ frontmatter: { date: DESC } }, { frontmatter: { title: DESC } }]
```
mirhamasala, line 10: What does this extra sort for the `title` do?

src/components/PostRelatedArticles/index.jsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+function PostRelatedArticles({ currentArticleTitle }) {
+  const data = useStaticQuery(query)
+
+  const currentArticleCategories = data.allMarkdownRemark.nodes
+    .filter((article) => article.frontmatter.title === currentArticleTitle)
+    .map((article) =>
+      article.frontmatter.categories?.map((category) => category.label)
+    )[0]
```
mirhamasala, line 46: Is it not possible to pass the category from the current article directly?

src/components/PostRelatedArticles/index.jsx
```diff
@@ -0,0 +1,71 @@
...
+        ?.map((category) => category.label)
+        .some((label) => currentArticleCategories?.includes(label))
+    )
+    .slice(0, 3)
+
+  return (
+    <div className={styles.root}>
+      {relatedArticles.length > 0 && (
+        <p className={styles.title}>Related Articles</p>
+      )}
```
mirhamasala, line 61: Why is the rest of the element returned if there are no related articles?

mirhamasala: And what about the parent styles?

```
<div className={styles.relatedArticles}>
    <PostRelatedArticles currentArticleTitle={title} />
</div>
 ```

Comment (mirhamasala) [reply src/components/PostRelatedArticles/index.jsx:61]: This may raise a "higher level" question ... what happens if there are no related articles? What do we display?

Comment (jreis13) [reply src/components/PostRelatedArticles/index.jsx:61]: If there are no articles they just don't render, but I agree I should move this check one level up. 

There's indeed the problem with the parent styles that I haven't been able to fix yet - when there are no articles I still have the spacing as if there were... will look into it

File: src/components/PostRelatedArticles/index.jsx [hunk 4]
```diff
@@ -0,0 +1,71 @@
...
+          title
+        }
+      }
+    }
+  }
+`
+
+function Article({ article }) {
+  return (
+    <div className={styles.article} key={article.fields.path}>
```
mirhamasala, line 30: I think we should make the whole card clickable

src/components/PostRelatedArticles/index.jsx
mirhamasala: Normally, this would be a question for Onda ... one thing we could do is display 3 random articles instead .. and instead of `Related`, it would be something like `More` - But I don't know if this is something we want. @gabrielpoca

src/components/PostRelatedArticles/index.jsx
mirhamasala: And we should probably use `Link` for navigating internal pages: https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/

src/components/PostRelatedArticles/index.module.scss
```diff
@@ -0,0 +1,57 @@
+@import "common/breakpoints";
+
+.root {
+  margin-bottom: 54px;
```
mirhamasala, line 4: Should be on the parent

src/components/PostRelatedArticles/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+            label
+          }
+          title
+        }
+      }
+    }
+  }
+`
+
+function Article({ article }) {
```
mirhamasala, line 29: This is a bit nitpicking. But we are using `Post` through the project. Introducing `Article` might be a bit confusing.

I would also call this a `PostCard` probably

src/components/PostRelatedArticles/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+  return (
+    <Link to={article.fields.path}>
+      <div className={styles.article} key={article.fields.path}>
+        <p className={styles.articleTitle}>{article.frontmatter.title}</p>
+        <p className={styles.articleLink}>Read more</p>
```
mirhamasala, line 34: Not an `articleLink` anymore.

src/components/PostRelatedArticles/index.jsx
mirhamasala: About the problem with the parent styles, if you have a layout element, that sets the spacing on `gap`, then you won't have this problem. (But let's pair today)

mirhamasala:
@jreis13 LGTM! 🌈👏🏻

PR #556, opened 2023-07-11
Show last podcasts

gabrielpoca:
It takes the 3 last episodes of each podcast and shows them in publishing order.

### DIFF
--- a/src/components/PodcastList/index.module.scss
+++ b/src/components/PodcastList/index.module.scss
@@ -12,14 +12,15 @@
 .item {
   display: flex;
   flex-direction: column;
-  padding-bottom: 84px;
   margin-right: 28px;
+  margin-bottom: 84px;
 }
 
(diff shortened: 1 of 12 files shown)


src/components/PodcastList/index.module.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.sectionTitle {
+  padding-top: 84px;
+  padding-bottom: 28px;
+}
+
+.item {
+  display: flex;
+  flex-direction: column;
+  padding-bottom: 84px;
```
joaop21, line 15: This bottom padding is clickable, which means if I click below the card, it will take me to its link.
The worst case is when I try to click the arrows and miss the click in the arrow. It will take me to the podcast page above the arrows.

Screenshot 2023-07-11 at 14 22 26

Issue #577, opened 2023-09-06
Create dedicated RSS feed


DavideSilva:
In order to submit the blog as a source of content to daily.dev, we need to create a dedicated RSS feed with only tech related content

PR #594, opened 2023-10-31
Add RSS feed for tech content

DavideSilva:
Why:
* #577

How:
* Adding a new feed to the `feeds` list that only fetches posts with the
  `engineering` category


gabrielpoca (approved):
👍 looks good

